PR #3, opened 2018-04-16
Add a simple example that select static data from a pg db

ChristophWurst:
A minimalist example that creates a new pool, runs `SELECT 1`
and prints the result.

Context: I'm currently having some issues [integrating this lib into Gotham](https://github.com/ChristophWurst/gotham-middleware-bb8) and so I wanted to write a minimalist bb8 program to verify this library actually works. I'm submitting this here because this might be useful for others exploring this lib.

### DIFF
--- a/examples/static_select.rs
+++ b/examples/static_select.rs
@@ -1,0 +1,42 @@
+extern crate bb8;
+extern crate bb8_postgres;
+extern crate futures;
+extern crate futures_state_stream;
+extern crate tokio_core;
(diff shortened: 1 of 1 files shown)


khuey:
Ok, great. Let me know if you need help figuring anything out.

khuey:
I moved this to the postgres folder to make it easier to build in 100af9fac8c25aa0e5c96715a29419d3e3203721

PR #8, opened 2018-08-29
use impl trait instead of boxing where possible

vorot93:


khuey:
Addressed in ef7763d2d34831d21448201abe6b9c229030e507

PR #10, opened 2018-11-14
Make the futures returned by bb8 `Send`

Daniel-B-Smith:
Without this, it is challening to use the connection pool with Hyper. Hyper expects all of its futures to be Send, so to use the pool, a Send future would need to wait on a non-Send future. That may be possible, but I have no idea how to do it.

Most of the changes are simply adding the annotations. The one substantial change is inside `get_idle_connection` where we have to send an `Arc<SharedPool>` unlocked and do the locking within each future. That change adds locking/unlocking both inside the `get_idle_connection` `loop_fn` and inside `run`.

There are some unrelated `rustfmt` changes because I have it tied to my editor's save. They are in isolated commits, so it is easy to see the diffs without them. I can revert them if you don't want to deal with them.

### DIFF
--- a/postgres/src/lib.rs
+++ b/postgres/src/lib.rs
@@ -53,7 +53,7 @@ impl bb8::ManageConnection for PostgresConnectionManager {
     fn is_valid
         (&self,
          conn: Self::Connection)
-         -> Box<Future<Item = Self::Connection, Error = (Self::Error, Self::Connection)>> {
+         -> Box<Future<Item = Self::Connection, Error = (Self::Error, Self::Connection)> + Send> {
         conn.batch_execute("")
     }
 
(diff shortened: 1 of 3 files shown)


Daniel-B-Smith:
Cleaner set of diffs without rustfmt: https://github.com/khuey/bb8/pull/10/files/ce5b2493ad2a9a1bd31576a20df244124f6d26b3..57924003c5ad501d48553e7245b4bff058acd20b

khuey:
Thank you for the PR. I'll take a look at it by the end of the week. The formatting changes are fine, you can leave those in.

Daniel-B-Smith:
Done. Thanks!

PR #11, opened 2018-11-14
Add connection pool for Redis

Daniel-B-Smith:
This creates a `RedisPool` and `RedisConnectionManager` type compatible with bb8. The `RedisPool` struct exists to paper over the small API incompatibility between the `redis` and `tokio-postgres` crates. While the `tokio-postgres` API always returns the connection, the `redis` API only returns the connection on success. The `RedisPool` hides that by creating a `RedisConnectionManager` that uses a `Option<Connection>` and returns `None` on failure. There's also a smaller incompatibility in that `tokio-postgres` uses `(T, Connection)` and `redis` uses `(Connection, T)`.

An example is forthcoming. I have used this code in a private project, so I just need to simplify the code to an example.

Depends upon https://github.com/khuey/bb8/pull/10

### DIFF
--- a/redis/Cargo.toml
+++ b/redis/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "bb8-redis"
+version = "0.1.0"
+authors = ["Daniel Smith <smith.daniel.br@gmail.com>", "Kyle Huey <khuey@kylehuey.com>"]
+
+[dependencies]
+bb8 = { path = ".." }
+futures = "0.1"
+tokio-core = "0.1"
+redis = "0.9"
--- a/redis/src/lib.rs
+++ b/redis/src/lib.rs
@@ -1,0 +1,99 @@
+//! Redis support for the `bb8` connection pool.
+#![deny(missing_docs, missing_debug_implementations)]
+
+pub extern crate bb8;
+pub extern crate redis;
+
+extern crate futures;
+extern crate tokio_core;
+
+use futures::{Future, IntoFuture};
+use tokio_core::reactor::Handle;
+
+use redis::async::Connection;
+use redis::{Client, RedisError};
+
+use std::io;
+use std::option::Option;
+
+type Result<T> = std::result::Result<T, RedisError>;
+
+/// `RedisPool` is a convenience wrapper around `bb8::Pool` that hides the fact that
+/// `RedisConnectionManager` uses an `Option<Connection>` to smooth over the API incompatibility.
+#[derive(Debug)]
+pub struct RedisPool {
+    pool: bb8::Pool<RedisConnectionManager>,
+}
+
+impl RedisPool {
+    /// Constructs a new `RedisPool`, see the `bb8::Builder` documentation for description of
+    /// parameters.
+    pub fn new(pool: bb8::Pool<RedisConnectionManager>) -> RedisPool {
+        RedisPool { pool }
+    }
+
+    /// Run the function with a connection provided by the pool.
+    pub fn run<'a, T, E, U, F>(&self, f: F) -> impl Future<Item = T, Error = E> + Send + 'a
+    where
+        F: FnOnce(Connection) -> U + Send + 'a,
+        U: IntoFuture<Item = (Connection, T), Error = E> + 'a,
+        U::Future: Send,
+        E: From<RedisError> + Send + 'a,
+        T: Send + 'a,
+    {
+        let f = move |conn: Option<Connection>| {
+            let conn = conn.unwrap();
+            f(conn)
+                .into_future()
+                .map(|(conn, item)| (item, Some(conn)))
+                .map_err(|err| (err, None))
+        };
+        self.pool.run(f)
+    }
+}
+
+/// A `bb8::ManageConnection` for `redis::async::Connection`s.
+#[derive(Clone, Debug)]
+pub struct RedisConnectionManager {
+    client: Client,
+}
+
+impl RedisConnectionManager {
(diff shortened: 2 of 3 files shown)


redis/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "bb8-redis"
+version = "0.1.0"
+authors = ["Kyle Huey <khuey@kylehuey.com>", "Daniel Smith <smith.daniel.br@gmail.com>"]
```
khuey: Put your name first please ;)

redis/Cargo.toml
```diff
@@ -0,0 +1,10 @@
...
+
+[dependencies]
+bb8 = { path = ".." }
+futures = "0.1"
+tokio-core = "0.1"
+redis = "0.9"
```
khuey: newline at EOF please

redis/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+use std::io;
+use std::option::Option;
+
+type Result<T> = std::result::Result<T, RedisError>;
+
+/// `RedisPool` is a convenience wrapper around `bb8::Pool` that hides the fact that
+/// `RedisConnectionManager` uses an `Option<Connection>` to smooth over the API incompatibility.
+#[derive(Debug)]
+pub struct RedisPool {
+    // The `bb8::Builder` that manages the pool.
```
khuey: Just "the underlying `bb8::Pool`" would be fine, though given that this is obvious and private idk that it's worth commenting.

redis/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+/// A `bb8::ManageConnection` for `redis::async::Connection`s.
+#[derive(Clone, Debug)]
+pub struct RedisConnectionManager {
+    client: Client,
+}
+
+impl RedisConnectionManager {
+    /// Create a new `PostgresConnectionManager`.
```
khuey: `RedisConnectionManager`

redis/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub fn run<'a, T, E, U, F>(&self, f: F) -> Box<Future<Item = T, Error = E> + Send + 'a>
+    where
+        F: FnOnce(Connection) -> U + Send + 'a,
+        U: IntoFuture<Item = (Connection, T), Error = E> + 'a,
+        U::Future: Send + 'a,
+        E: From<RedisError> + Send + 'a,
+        T: Send + 'a,
+    {
+        let f = move |conn: Option<Connection>| {
+            let conn = conn.unwrap();
```
khuey: Wouldn't it make more sense to throw an error here than unwrap (and potentially panic)?

redis/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    fn is_valid(
+        &self,
+        conn: Self::Connection,
+    ) -> Box<Future<Item = Self::Connection, Error = (Self::Error, Self::Connection)> + Send> {
+        // The connection should only be None after a failure.
+        Box::new(
+            redis::cmd("PING")
+                .query_async(conn.unwrap())
```
khuey: Similarly here, if the conn is `None` that should result in returning an error, not panicking.

redis/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+        U::Future: Send + 'a,
+        E: From<RedisError> + Send + 'a,
+        T: Send + 'a,
+    {
+        let f = move |conn: Option<Connection>| {
+            let conn = conn.unwrap();
+            f(conn)
+                .into_future()
+                .map(|(conn, item)| (item, Some(conn)))
+                .map_err(|err| (err, None))
```
khuey: I don't like that on error the underlying connection is thrown away here. This precludes ever usefully implementing an equivalent to the `CustomizeConnection::on_release` functionality in r2d2. Is there a reason not to return the connection in the error case?

Daniel-B-Smith: Unfortunately, the underlying library swallows the connection on error:

https://docs.rs/redis/0.9.1/redis/struct.Cmd.html#method.query_async
https://docs.rs/redis/0.9.1/redis/type.RedisFuture.html

redis/src/lib.rs
Daniel-B-Smith: Done.

redis/src/lib.rs
Daniel-B-Smith: I may have misunderstood the invariants provided by the bb8 library. My understanding was that this method would never get called with a connection after an error. Unwrapping both makes the code simpler and effectively serves as an assert. This will only ever be `None` if there was already a failed call to `run` or `is_valid`. I can change it to be an error if my understanding is incorrect or you'd rather we return an error regardless.

khuey: In order to get here the connection must be either a) freshly generated from `add_connection` or b) returned from a previous `Pool::run` call. In the first case, `RedisConnectionManager::connect` always returns `Some`, and in the second case `RedisConnectionManager::has_broken` would have rejected any connection that is `None` before we put it back in the pool, so I believe you are correct here.

redis/src/lib.rs
khuey: Hrm ok :/

khuey:
Thanks!

apiraino:
@Daniel-B-Smith (or @khuey ) thanks this PR. Would you be able to provide the usage examples you mentioned? I was investigating bb8, too. thanks again :+1:

PR #14, opened 2019-03-03
Fix redis dependency version

vorot93:
Actually, after additional testing, it appears that neither caret nor tilde play well with 0.x versions. Therefore we must use `>=` to remain compatible with both redis 0.9 and 0.10.


khuey:
FYI @Daniel-B-Smith

khuey:
Thanks.

PR #16, opened 2019-03-13
minor tweaks

jspeis:
I'm a bit new to Rust in general but here are two changes I've made locally that might also be helpful to others

* fixes type in example
* adds the `Clone` trait for  `PostgresConnectionManager`


khuey:
Looks great, thanks!

PR #21, opened 2019-05-16
Issue #18 Implement `Error` for `RunError`

### LINKED ISSUE #18: Implement `Error` for `bb8::RunError`
Hello, thanks for the library. Is it possible to implement [the Error](https://doc.rust-lang.org/nightly/std/error/trait.Error.html) trait for the [RunError](https://docs.rs/bb8/0.3.0/bb8/enum.RunError.html)? This would allow better cooperation with the `failure` crate.

elpiel:
Closes #18 

Not sure if this is how it's supposed to be implemented, but better to open the PR and get some feedback on it.

Maybe also a clue how to write a Unit test or add a test to the one existing to check the valid implementation.


khuey:
It looks pretty good.

src/lib.rs
```diff
@@ -62,6 +63,28 @@ pub enum RunError<E> {
     TimedOut,
 }
 
+impl<E> fmt::Display for RunError<E>
+    where E: error::Error + 'static
+{
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            RunError::User(ref err) => write!(f, "User error: {}", err),
+            RunError::TimedOut => write!(f, "Timed Out"),
```
khuey: I think it would be better to pass through the user error as is and then change timeout to something like "Timed out in bb8"

elpiel:
@khuey thanks for checking! Can you see the latest changes. I also thing that we might pass it like that. Looks better.

elpiel:
@khuey ping?

src/lib.rs
```diff
@@ -62,6 +63,28 @@ pub enum RunError<E> {
     TimedOut,
 }
 
+impl<E> fmt::Display for RunError<E>
+    where E: error::Error + 'static
+{
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            RunError::User(ref err) => fmt::Display::fmt(&err, f),
```
Txuritan: Just a suggestion but this looks cleaner in my book.

```suggestion
            RunError::User(ref err) => err.fmt(f),
```

elpiel:
Thanks for the suggestion @Txuritan !

khuey:
Sorry for dropping the ball here. It looks great. Thanks to you too @Txuritan.

PR #23, opened 2019-06-19
Added example for async/await/futures0.3 usage to .run()

kaibyao:
Ran into this for a few hours today, couldn't figure out how to convert an async/await future to work with `client.run()`. Figured this would help others.


khuey:
Thanks!

PR #25, opened 2019-07-19
Fix broken build

evdokimovs:
You specified dependency like this `redis = ">=0.9"`. Because of that, cargo automatically updated `redis` version to `0.11` in which some [breaking changes] were made. The name of the module `async` was renamed to `aio`.

Also `master` branch don't compile because of a error in `src/lib.rs`. This error I fixed too.

[breaking changes]: https://github.com/mitsuhiko/redis-rs/blob/master/CHANGELOG.md#breaking-changes


khuey:
Thanks!

PR #33, opened 2019-08-27
Update redis-rs to 0.12 in bb8-redis dependencies.

singulared:


redis/Cargo.toml
```diff
@@ -10,4 +10,4 @@ repository = "https://github.com/khuey/bb8"
 bb8 = { path = "../" }
 futures = "0.1"
 tokio = "0.1"
-redis = "0.11"
+redis = "0.12"
```
khuey: This requires bumping the version of bb8-redis as well to 0.4.0-pre, since the new redis is semver incompatible.

singulared: Yep, I think you are right. Is this a problem?

khuey: No, it's not a problem.

khuey:
Thanks!

PR #40, opened 2019-11-01
Expose underlying bb8::Pool from RedisPool

bacek:
Made `RedisPool.pool` public to access `bb8::Pool.stats()`


khuey:
Thanks. I've pushed a different approach that doesn't allow `RedisPool` to be directly constructed by library users but it should be just as good for your use case.

PR #45, opened 2019-11-27
Drop based interface

### LINKED ISSUE #19: A way to not have `run()` with a closure?
First of all - sorry for my ignorance, I'm new to Rust.
In the project I am working we want to use bb8 and we're successfully using it, but something like that:
```
let fut = self.db_pool
            .run(move |mut conn| {
                conn.prepare("SELECT channel_id, creator, deposit_asset, deposit_amount, valid_until FROM channels")
                    .then(move |res| match res {
                        Ok(stmt) => {
                            conn
                                .query(&stmt, &[])
                                .collect()
                                .into_future()
                                .then(|res| match res {
                                    Ok(rows) => Ok((rows, conn)),
                                    Err(err) => Err((err, conn)),
                                })
                                .into()
                        }
                        Err(err) => try_future!(Err((err, conn))),
                    })
                    .and_then(|(rows, conn)| {
                        let channels = rows.iter().map(|row| {
                            Channel {
                                id: row.get("channel_id"),
                                creator: row.get("creator"),
                                deposit_asset: row.get("deposit_asset"),
                                deposit_amount: row.get("deposit_amount"),
                                valid_until: row.get("valid_until"),
                            }
                        }).collect();

                        Ok((channels, conn))
                    })
            })
            .map_err(|err| handle_internal_error(&err));
```

Looks quite, well - unbreakable. + we cannot use `await!()` inside the closure.

From my understanding some things are happening in `run()` that handle the dropping/idling of the connection. Isn't it possible that this can be change in such a way that `Drop` (or tbh I don't know what exactly) is implemented that `run()` can return the connection and we can use it with `await!()`, but also as a future and to get rid of the tidies `Ok((channels, conn))` returns?

Thank you in advance! I will be happy to help and implement this as long as I understand how.

elpiel:
Closes #19 

This is the follow up of https://github.com/djc/bb8/pull/3
I've combined all changes into 1 commit.
Thank you @djc for your help!
@khuey this help a lot with how this crate is used and I need some feedback in order to finish it up.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -102,7 +102,7 @@ impl<E> ErrorSink<E> for NopErrorSink {
     fn sink(&self, _: E) {}
 
     fn boxed_clone(&self) -> Box<dyn ErrorSink<E>> {
-        Box::new(self.clone())
+        Box::new(*self)
     }
 }
 
@@ -148,7 +148,7 @@ where
     fn make_idle(conn: Conn<C>) -> IdleConn<C> {
         let now = Instant::now();
         IdleConn {
-            conn: conn,
+            conn,
             idle_start: now,
         }
     }
@@ -451,10 +451,7 @@ where
         Ok(conn) => {
             let now = Instant::now();
             let conn = IdleConn {
-                conn: Conn {
-                    conn: conn,
-                    birth: now,
-                },
+                conn: Conn { conn, birth: now },
                 idle_start: now,
             };
 
@@ -602,7 +599,7 @@ impl<M: ManageConnection> Pool<M> {
 
         let shared = Arc::new(SharedPool {
             statics: builder,
-            manager: manager,
+            manager,
             internals: Mutex::new(internals),
         });
 
@@ -716,7 +713,7 @@ impl<M: ManageConnection> Pool<M> {
         let birth = conn.birth;
         let (r, mut conn): (Result<_, E>, _) = match f(conn.conn).await {
             Ok((t, conn)) => (Ok(t), conn),
-            Err((e, conn)) => (Err(e.into()), conn),
+            Err((e, conn)) => (Err(e), conn),
         };
 
         // Supposed to be fast, but do it before locking anyways.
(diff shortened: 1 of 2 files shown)


Xorlev:
I gave this a spin since I found myself wanting similar semantics for a fork of the sidekiq crate which uses async-await + bb8.

I think this needs an implementation of `DerefMut` for `PooledConnection`, otherwise it's (seemingly) not possible to use with any library expecting a mut ref to an underlying connection, such as Redis.

FWIW, r2d2 has [an implementation of DerefMut](https://docs.rs/r2d2/0.8.6/src/r2d2/lib.rs.html#631-638) as well.

elpiel:
Good suggestion, I've implemented it!

src/lib.rs
```diff
@@ -725,16 +733,70 @@ impl<M: ManageConnection> Pool<M> {
...
 
+    /// Return connection back in to the pool
+    fn put_back(&self, birth: Instant, mut conn: Conn<M::Connection>) {
+        let inner = self.inner.clone();
+
+        // Supposed to be fast, but do it before locking anyways.
+        let broken = inner.manager.has_broken(&mut conn.conn);
+
+        futures::executor::block_on(async {
+            let mut locked = inner.internals.lock().await;
```
djc: It looks like this code is duplicated from `run()`. It should probably be factored out so that `run()` and `put_back()` can share as much logic as possible. (Probably better to move the `broken` value inside the async block, too.)

src/lib.rs
```diff
@@ -725,16 +733,70 @@ impl<M: ManageConnection> Pool<M> {
...
+                locked.put_idle_conn(conn);
+            }
+        })
+    }
+
+    /// Retrieves a connection from the pool.
+    pub async fn get(&self) -> Result<PooledConnection<M>, RunError<M::Error>> {
+        let inner = self.inner.clone();
+
+        match get_idle_connection(inner).await {
```
djc: Same thing here: this code should be factored out so that `run()` and `get()` can share the same code path.

djc:
I'd like to see this as three separate commits, rebased on current master:

* Clean up clippy problems
* Factor out the "get connection" and "put back connection" into separate methods
* Implemented `PooledConnection`

(The latter two should of course not reintroduce any formatting/clippy issues.)

Thanks for working on this!

elpiel:
I've managed to break one of the tests. It now tries to `unwrap()` on a `RunError::Timeout` error.
I will need to look into this.

djc:
Do note that one of the tests fails on master anyway, so this might not be caused by your PR.

elpiel:
Then it's ready for review @djc

PR #51, opened 2020-01-01
Retry connection establishment, defaulting to 3 tries

djc:
This fixes the test from #45. @bbigras, would you like to review? You managed to find the one `TODO` in the library. @khuey, if you have some time to spare, review would be much appreciated too. I'd like to get this merged and then put out a first release soon after.

### DIFF
--- a/bb8/src/lib.rs
+++ b/bb8/src/lib.rs
@@ -27,7 +27,7 @@ use futures::lock::{Mutex, MutexGuard};
 use futures::prelude::*;
 use futures::stream::FuturesUnordered;
 use tokio::spawn;
-use tokio::time::{interval_at, timeout, Interval};
+use tokio::time::{delay_for, interval_at, timeout, Interval};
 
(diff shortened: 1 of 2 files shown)


djc:
For easiest review, [ignore the whitespace changes](https://github.com/khuey/bb8/pull/51/commits/fa98aabb88ba47221c6fef94ea4e164294a2313c?w=1).

djc:
Points of potential discussion:

* Bikeshedding the configuration item name
* Is the negative test useful here?

djc:
I changed this to not rely on a number of retries and instead copy the exponential backoff strategy from r2d2. However, for now it will terminate after `connection_timeout`. I think this is a pretty good solution for now and will merge this, happy to get any feedback post-merge.

bbigras:
Thanks for the work @djc . Sorry I didn't reply last time. I have been busy and lazy during the holidays.

djc:
@bbigras, no problem, that's what holidays are for!

PR #56, opened 2020-01-17
Adding actix example for postgres adapter

friktor:
@djc added an example identical to hyper, uses the same db docker container.

### DIFF
--- a/postgres/Cargo.toml
+++ b/postgres/Cargo.toml
@@ -24,4 +24,6 @@ tokio-postgres = "0.5.1"
 
 [dev-dependencies]
 hyper = { version = "0.13" }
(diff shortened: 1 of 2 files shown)


djc:
So is this working for you now?

friktor:
@djc no unfortunately. But I decided to add it as a test example to the repository, because the actix is now more relevant than hyper. On it you can verify the library worked, and as an example for those who use actix. To my regret, I am very poorly versed in tokio, futures and cannot help with fixing the main error.

djc:
Given the current situation with Actix and the amount of dependencies this pulls in, I don't think it's a good fit. I'm even on the fence about the hyper example. Sorry!

PR #58, opened 2020-03-11
Use connection info for connection manager

pksunkara:
This would be breaking change. We will document it correctly but it will be better for the future.

### DIFF
--- a/redis/src/lib.rs
+++ b/redis/src/lib.rs
@@ -8,7 +8,7 @@ use futures::future::{Future, FutureExt};
 
 use async_trait::async_trait;
 use redis::aio::Connection;
-use redis::{Client, RedisError};
+use redis::{Client, IntoConnectionInfo, RedisError};
 
 /// `RedisPool` is a convenience wrapper around `bb8::Pool` that hides the fact that
 /// `RedisConnectionManager` uses an `Option<Connection>` to smooth over the API incompatibility.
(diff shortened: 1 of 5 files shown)


djc:
It's not clear to me why this is an improvement. What are you trying to achieve here?

pksunkara:
Simplicity in using the library.

djc:
How is this simpler? It's not obvious to me. (Also note that CI is failing.)

pksunkara:
Before this PR the usage is:

```rust
let manager = RedisConnectionManager::new(Client::open(env("REDIS_URL")).unwrap()).unwrap();
```

After this PR the usage is:

```rust
RedisConnectionManager::new(env("REDIS_URL")).unwrap();
```

pksunkara:
I have fixed the tests.

redis/src/lib.rs
```diff
@@ -51,13 +51,13 @@ impl RedisPool {
...
 pub struct RedisConnectionManager {
-    client: Client,
+    connection_info: ConnectionInfo,
 }
 
 impl RedisConnectionManager {
     /// Create a new `RedisConnectionManager`.
-    pub fn new(client: Client) -> Result<RedisConnectionManager, RedisError> {
-        Ok(RedisConnectionManager { client })
+    pub fn new<T: IntoConnectionInfo>(info: T) -> Result<RedisConnectionManager, RedisError> {
```
djc: I think we should change `new()` to build the `Client` from `T: IntoConnectionInfo` here, so that we don't have to build a `Client` for every `connect()` call.

djc:
Thanks!

Issue #63, opened 2020-04-26
Change `is_valid()` to take a mutable Self::Connection, instead of accepting then returning an owned value?


naerbnic:
It's conceivable that, even in a case where a connection is no longer valid, that other parts of the connection manager may want to access the connection value. For instance, the r2d2 CustomizeConnection trait will pass that to on_release(), which can perform more custom operations on it. Is there any reason, aside from backwards compatibility, that `is_valid()` does not take Self::Connection by mutable reference, then have the Result be `Result<(), Self::Error>`? I believe this would better deal with the circumstances, and still permit the connection itself to be dropped.

djc:
Yeah, I think that would make sense. Would you be interested in implementing this?

naerbnic:
Certainly. I have a version ready to go. I'll send it over shortly.

djc:
Fixed in #64.

PR #64, opened 2020-04-28
Is valid takes mut reference

naerbnic:
Change `is_valid()` to take the connection by reference. This simplifies the API, and lets the resulting connection be kept on error, allowing for features like CustomizeConnection to be implemented in the future. #63


djc:
I think this looks good. Thanks!

PR #69, opened 2020-05-21
Remove obsolete Option wrapper for Redis connections

djc:


seunlanlege:
this change is very much needed :grinning:

PR #71, opened 2020-05-23
Add redis pubsub pool and examples

sync:
The Redis crate recently added redis pubsub subscription and would like to use this feature with bb8.

https://github.com/mitsuhiko/redis-rs/blob/master/examples/async-pub-sub.rs

PS: I am still super new to rust so probably can be cleaned up

### DIFF
--- a/redis/Cargo.toml
+++ b/redis/Cargo.toml
@@ -15,3 +15,4 @@ redis = "0.16"
 
 [dev-dependencies]
 tokio = { version = "0.2", features = ["macros"] }
+futures-util = "0.3"
--- a/redis/src/lib.rs
+++ b/redis/src/lib.rs
@@ -39,7 +39,7 @@ pub use bb8;
 pub use redis;
 
 use async_trait::async_trait;
-use redis::aio::Connection;
+use redis::aio::{Connection, PubSub};
 use redis::{Client, IntoConnectionInfo, RedisError};
 
 /// A `bb8::ManageConnection` for `redis::Client::get_async_connection`.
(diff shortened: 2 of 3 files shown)


redis/Cargo.toml
```diff
@@ -15,3 +15,4 @@ redis = "0.16"
 
 [dev-dependencies]
 tokio = { version = "0.2", features = ["macros"] }
+futures-util = "0.3"
```
sync, line 18: this is just need for the new redis examples, let me know if there is a better way of doing this ?

redis/src/lib.rs
```diff
@@ -75,3 +75,42 @@ impl bb8::ManageConnection for RedisConnectionManager {
...
+    pub fn new<T: IntoConnectionInfo>(info: T) -> Result<RedisPubSubConnectionManager, RedisError> {
+        Ok(RedisPubSubConnectionManager {
+            client: Client::open(info.into_connection_info()?)?,
+        })
+    }
+}
+
+#[async_trait]
+impl bb8::ManageConnection for RedisPubSubConnectionManager {
+    type Connection = PubSub;
```
sync, line 97: need to do 2 things, change connection type to `PubSub`

redis/src/lib.rs
```diff
@@ -75,3 +75,42 @@ impl bb8::ManageConnection for RedisConnectionManager {
...
+    type Error = RedisError;
+
+    async fn connect(&self) -> Result<Self::Connection, Self::Error> {
+        self.client
+            .get_async_connection()
+            .await
+            .map(|conn| conn.into_pubsub())
```
sync, line 104: turn async connection into a pusbub connection `into_pusbsub`

redis/src/lib.rs
```diff
@@ -75,3 +75,42 @@ impl bb8::ManageConnection for RedisConnectionManager {
...
+    }
+
+    async fn is_valid(&self, conn: &mut Self::Connection) -> Result<(), Self::Error> {
+        let test_sub_name = "__test__-sub-from-bb8";
+        conn.subscribe(test_sub_name).await?;
+        conn.unsubscribe(test_sub_name).await
```
sync, line 110: this feels hackish, but there is not much that can be done on a pubsub connection, basic CMD aren't available for example

djc:
I'm not really comfortable with supporting a pubsub pool as part of bb8-redis. It seems like a relatively niche use case and it's not at all clear that (most usages of) pubsub connections are a good match for pooling because the length of the "loan" from the pool could be relatively high. I think `dedicated_connection()` is a better match, as @khuey suggested. If you want a pubsub pool, you can of course always maintain one inside your application or as part of a separate crate.

sync:
not a problem and thanks for the suggestion

PR #77, opened 2020-08-11
Disable default features for redis crate

x04:
Since this library already depends on tokio I see no reason for the async-std feature of the redis crate to be enabled. It adds ~30 extra dependencies that have to be built even though the feature should never be used.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 **/*.rs.bk
 Cargo.lock
 *~
+/.idea/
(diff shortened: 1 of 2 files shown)


djc:
I think we should only enable the features that are needed to bring in the interfaces we use, not any other features (those can be enabled by a downstream crate). Also, please remove the unrelated .gitignore changes.

djc:
Thanks!

PR #80, opened 2020-10-03
Add support for lapin

adrianbenavides:
This MR adds support for [lapin](https://github.com/CleverCloud/lapin).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,6 +21,12 @@ jobs:
 
     steps:
       - uses: actions/checkout@v1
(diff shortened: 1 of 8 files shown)


djc:
Is there a particular reason for adding the lapin support to this repo, rather than just publishing it as a separate crate? I don't think it necessarily makes sense to integrate it here, as I (as the most active maintainer -- even though I haven't been very active lately) won't have much time or energy to work on this.

adrianbenavides:
Yeah that makes totally sense. The project structure led me to think that the available support crates were included here, as opposed to r2d2.

Thanks @djc for your time!

PR #89, opened 2020-12-12
Adds feature for users to customize connections created by the pool

agersant:
Implementation for https://github.com/djc/bb8/issues/88

I ended up running into not too many issues!

### DIFF
--- a/bb8/src/api.rs
+++ b/bb8/src/api.rs
@@ -207,8 +207,8 @@ impl<M: ManageConnection> Builder<M> {
         self
     }
 
-    /// Set the connection customizer which will be used to initialize and cleanup
-    /// connections created and destroyed by the pool.
+    /// Set the connection customizer which will be used to initialize
+    /// connections created by the pool.
     ///
     /// Defaults to `NopConnectionCustomizer`.
     pub fn connection_customizer(
--- a/bb8/src/api.rs
+++ b/bb8/src/api.rs
@@ -84,7 +84,7 @@ pub struct Builder<M: ManageConnection> {
     pub(crate) error_sink: Box<dyn ErrorSink<M::Error>>,
     /// The time interval used to wake up and reap connections.
     pub(crate) reaper_rate: Duration,
-    /// User-supplied trait object responsible for initializing and cleaning-up connections
+    /// User-supplied trait object responsible for initializing connections
     pub(crate) connection_customizer: Box<dyn CustomizeConnection<M::Connection, M::Error>>,
     _p: PhantomData<M>,
 }
@@ -268,7 +268,7 @@ pub trait ManageConnection: Sized + Send + Sync + 'static {
     fn has_broken(&self, conn: &mut Self::Connection) -> bool;
 }
 
-/// A trait which provides functionality to initialize and cleanup a connection
+/// A trait which provides functionality to initialize a connection
 #[async_trait]
 pub trait CustomizeConnection<C: Send + 'static, E: 'static>:
     std::fmt::Debug + Send + Sync + 'static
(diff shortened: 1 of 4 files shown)


bb8/src/inner.rs
```diff
@@ -194,6 +209,22 @@ where
             }
         }
     }
+
+    async fn on_acquire_connection(
+        &self,
+        mut conn: M::Connection,
```
djc, line 215: Should this take a `&mut` as well, rather than taking ownership? The semantics here can be tricky since a future might get dropped during an await point, in which case we'd leak the connection here.

djc:
Overall this is looking good, thanks for working on it!

bb8/src/inner.rs
agersant: That was my first choice too but I couldn't get it to compile that way (and the semantics nuance is a bit lost on me). 

Here is what it looks like:
https://github.com/agersant/bb8/commit/2fd5045e5a49c589047409df4de92bf8b8701c86
(or [compacted](https://github.com/agersant/bb8/commit/50945f6f5a1fee914cd0fb2c50852b577548fce1) with `and_then`)

The borrow checker is not happy with either of these and I'm not sure how to fix that.

image

bb8/src/api.rs
```diff
@@ -253,6 +268,38 @@ pub trait ManageConnection: Sized + Send + Sync + 'static {
...
+    }
+
+    /// Called with connections when they are removed from the pool.
+    ///
+    /// The connections may be broken (as reported by `ManageConnection::is_valid` or
+    /// `ManageConnection::has_broken`), or have simply timed out.
+    ///
+    /// The default implementation does nothing.
+    #[allow(unused_variables)]
+    fn on_release(&self, connection: &mut C) {}
```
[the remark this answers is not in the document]
agersant: Yeah that makes sense. I gave that a shot and also made it return a `Result<(), E>` for consistency.

https://github.com/agersant/bb8/commit/5d5ff3210299d6838b75a11c5cbac0d5dcd0bca6

Unfortunately, this breaks two tests (annotated). I think none of this would be a problem if `put_back` was async, but it's called from `PooledConnection::drop` so that doesn't seem possible. Unsure where to go from there :(

bb8/src/inner.rs
djc: Can you try writing it like this?

```rust
let conn = self.inner.manager.connect().await;
self.on_acquire_connection(&mut conn).await?;
Ok(conn)
```

bb8/src/api.rs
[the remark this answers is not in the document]
djc: Hmm, that is annoying. Do you actually need `on_release()` for your use case? If not, I suggest we just make do with `on_acquire()` for this PR.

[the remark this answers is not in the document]
agersant: Works for me!

bb8/src/inner.rs
agersant: Haha, that works fine. Less is more!

The other call site looks a little less nice, but I got it to compile but just manually unrolling the `map` logic.

agersant:
I think this is ready for another look. Removing `on_release` made it a much smaller PR too.

bb8/src/inner.rs
```diff
@@ -174,7 +176,14 @@ where
...
         let mut delay = Duration::from_secs(0);
         loop {
-            match shared.manager.connect().await {
+            let conn = match shared.manager.connect().await {
+                Err(e) => Err(e),
+                Ok(mut c) => match self.on_acquire_connection(&mut c).await {
+                    Err(e) => Err(e),
+                    Ok(_) => Ok(c),
+                },
+            };
```
djc, line 185: The repeated `Err(e) => Err(e)` arms can probably be factored out by using some combinators, can you look into that? And while you're in there, please add an empty line between first `let conn = match` and `match conn {` blocks?

djc:
This is looking good, some minor style things and then I think this is good to merge!

bb8/src/inner.rs
agersant: Updated. Looks a bit more compact.

djc:
Thanks!

PR #94, opened 2020-12-25
CustomizeConnection::on_release()

djc:
@agersant how do you feel about this?

### DIFF
--- a/bb8/src/api.rs
+++ b/bb8/src/api.rs
@@ -273,13 +273,18 @@ pub trait CustomizeConnection<C: Send + 'static, E: 'static>:
     /// Called with connections immediately after they are returned from
     /// `ManageConnection::connect`.
     ///
-    /// The default implementation simply returns `Ok(())`.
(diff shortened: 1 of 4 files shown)


agersant (approved):

agersant:
Looks good to me, thanks for following up on this!

djc:
I ended up removing it before release, since I'm not confident in the mechanics (I found at least one gap which wasn't that easy/clean to solve), and I can't really think of a use case. Will leave this out for now and add it back if someone comes up with a use case.

paolobarbolini:
This is exactly what we need for `lettre`. We have a PR  https://github.com/lettre/lettre/pull/550 for adding a connection pool to our async version of the API, but we need a way for [quit](https://github.com/lettre/lettre/blob/bbf56de83d20e60187cc95a81c11c7f019a33da6/src/transport/smtp/client/async_connection.rs#L201) to be called before dropping a connection.

djc:
Can you open a new issue and preferably a PR?

PR #96, opened 2021-02-09
Fix ping request

invis87:
Hopefully this fix https://github.com/djc/bb8/issues/95
On my Dev errors dissapeared


djc:
Okay, please dump all the `Cargo.toml` changes and only keep the changes in `redis/src/lib.rs`; all the other changes are unnecessary (you should be able to update tokio and futures by running `cargo update` in your project).

invis87:
Ok, I will create another branch.

Issue #73, opened 2020-07-10
Postgres: DISCARD ALL


jakajancar:
Would be nice if bb8-postgres supported calling [`DISCARD ALL`](https://www.postgresql.org/docs/12/sql-discard.html) after the connection is returned to the pool, so it's guaranteed to be "pristine" on the next checkout.

Unless I'm missing something, there isn't even a "post check-in" hook in bb8 right now, only `test_on_check_out`. I suppose that could be used, but the best is to do it early, release the resources (e.g. `UNLISTEN`) and reduce the latency on checkout (assuming `test_on_checkout` is false).

djc:
Might be nice to have a look at other pooling solutions (r2d2, deadpool, sqlx) and see if/how they support this.

(Note that previously we recommended that LISTEN queries should probably be issued on non-pooled connections, through `Pool::dedicated_connection()`.)

jakajancar:
PgBouncer has it configurable (‘ server_reset_query’) but it defaults to ‘DISCARD ALL’. PgBouncer is probably the gold standard when it comes to Postgres pooling.

It’s really not about the LISTEN, but all (currently 9) resources this releases. Another nice thing is that it fails if the connection is inside a transaction, so it helps automatically check that you’re closing them correctly.

djc:
I was referring mostly to Rust crates.

I'm unlikely to have time for this any time soon unless someone is willing to pay for it. Will review hooks for this in the connection manager trait, though.

jakajancar:
Yes, I think hooks are key rather than supporting every single possible reset query. Then one can subclass the postgres manager and do what they want.

djc:
#89 added a way to customize connections at the start of the lifetime rather than the end, which should help with this.

ErikDeSmedt:
I know this issue is old but I've just been bitten by this and would be interested in seeing a fixx

djc:
Happy to review a PR for this; I don't currently have Postgres set up so won't be working on this unless someone wants to fund it.

(If someone decides to use an LLM: you're expected to self-review and edit before submitting, and write the PR description and comments yourself.)

PR #100, opened 2021-03-14
Add new postgres manager builder and support a custom validation query

film42:
Issue #73 was looking to make a "DISCARD ALL" query possible when a
connection is returned to the pool. Because async drop isn't a thing,
this isn't really possible at the moment. However, we can use the
`is_valid` method on the `ManagedConnection` trait to discard all
session state before yielding the connection.

This PR makes it possible to change the query used within the `is_valid`
method call in `bb8_postgres`.

To make configuring the validation query a little easier, I added a new
postgres connection manager builder helper type. Let me know if you're
:+1: or :-1: on that.

This also includes an example called buidler.rs that shows how someone
could use `DISCARD ALL` as a validation query. It then prints a few
things to show that session state was indeed cleared between checkouts.

```
$ cargo run --example builder
...
The current connection PID: 86445
BB8 says, "beep boop"
The current connection PID: 86445
After DISCARD ALL on checkout, BB8 says, ""
```

### DIFF
--- a/postgres/examples/builder.rs
+++ b/postgres/examples/builder.rs
@@ -1,0 +1,78 @@
+use bb8::{Pool, PooledConnection, RunError};
+use bb8_postgres::{PostgresConnectionManager, PostgresConnectionManagerBuilder};
+
+use std::error::Error;
+use std::str::FromStr;
+
+// Demonstrate the postgres connection manager builder and its validation query.
+//
+// The simplest way to start the db is using Docker:
+// docker run --name gotham-middleware-postgres -e POSTGRES_PASSWORD=mysecretpassword -p 5432:5432 -d postgres
+#[tokio::main]
+async fn main() -> Result<(), Box<dyn Error>> {
+    let config =
+        tokio_postgres::config::Config::from_str("postgresql://postgres:docker@localhost:5432")
+            .unwrap();
+
+    let pg_mgr = PostgresConnectionManagerBuilder::new()
+        .tls(tokio_postgres::NoTls)
+        .validation_query("DISCARD ALL".into())
+        .config(config)
+        .build()?;
+
+    let pool = match Pool::builder().build(pg_mgr).await {
+        Ok(pool) => pool,
+        Err(e) => panic!("builder error: {:?}", e),
+    };
+
+    // The first checkout sets a session parameter.
+    {
+        let connection = pool.get().await?;
+        println!(
+            "The current connection PID: {}",
+            get_pid(&connection).await?
+        );
+        set_session_parameter(&connection, "bb8.greeting", "beep boop").await?;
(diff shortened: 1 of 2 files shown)


djc:
Are there any reasons not to use [`CustomizeConnection`](https://docs.rs/bb8/0.7.0/bb8/trait.CustomizeConnection.html) for this? It seems like a better fit than shoehorning stuff into `is_valid()`.

film42:
Extending customized connection is certainly doable. I attached a diff below that adds an `on_check_out` callback. With this you could make `DISCARD ALL` work with bb8_postgres by adding a `DISCARD ALL` query to your customized connection's `on_check_out` impl. You would probably want to disable `test_on_check_out` because the `on_check_out` call is essentially doing the same thing. I like that the callback makes bb8 more flexible, but it makes the postgres wrapper a little strange. Postgres should probably be responsible for customizing what runs under `is_valid`, but the customized connection route is a much smaller change! I'm happy either way.

```diff
diff --git a/bb8/src/api.rs b/bb8/src/api.rs
index 77bf219..2a3e788 100644
--- a/bb8/src/api.rs
+++ b/bb8/src/api.rs
@@ -278,6 +278,12 @@ pub trait CustomizeConnection<C: Send + 'static, E: 'static>:
     async fn on_acquire(&self, _connection: &mut C) -> Result<(), E> {
         Ok(())
     }
+
+    /// Called immediately before the connection has been checked out. This is called
+    /// before [`ManageConnection::is_valid`] is called.
+    async fn on_check_out(&self, _connection: &mut C) -> Result<(), E> {
+        Ok(())
+    }
 }
 
 /// A smart pointer wrapping a connection.
diff --git a/bb8/src/inner.rs b/bb8/src/inner.rs
index 4c20e42..e1a0a0e 100644
--- a/bb8/src/inner.rs
+++ b/bb8/src/inner.rs
@@ -95,6 +95,12 @@ where
                 }
             };
 
+            if let Err(e) = self.on_check_out_connection(&mut conn).await {
+                self.inner.statics.error_sink.sink(e);
+                conn.drop_invalid();
+                continue;
+            }
```

djc:
Yeah, I think the `on_check_out()` hook in `CustomizeConnection` is right for this. If you like, we can add a `CustomizeConnection` impl that implements the `DISCARD ALL` behavior to bb8-postgres.

djc:
Although thinking about this more, it's probably more correct to do this on check in? Since it really doesn't need to happen to new connections, and also doing it on check out increases the latency for the checkout path. That just gets into the problem of running async code in `Drop` again though.

film42:
@djc Running this on check in is certainly preferred but becomes an async drop problem. I'll play around some more, but the customized connection option is a good backup plan.

PR #104, opened 2021-03-30
Implement non-retriable errors

marcbowes:
`ManageConnection::should_retry` can now be implemented to prevent bb8
from attempting to retry connection attempts. This serves two purposes:

1. The caller of `pool.get` will now receive the underlying exception if
a "catastrophic" error is hit. Previously, the caller would get a
`TimedOut` with no cause.

2. The error can be surfaced quickly, rather than going through several
connection attempts.

This is particularly useful for interactive use of bb8, such as allowing
a user to directly input a connection to a program such a CLI. If the
connection is incorrect, it is desirable to fail fast with a useful
error message.

### DIFF
--- a/bb8/src/api.rs
+++ b/bb8/src/api.rs
@@ -255,7 +255,7 @@ pub trait ManageConnection: Sized + Send + Sync + 'static {
     /// The connection type this manager deals with.
     type Connection: Send + 'static;
     /// The error type returned by `Connection`s.
-    type Error: fmt::Debug + Send + 'static;
+    type Error: fmt::Debug + Send + Clone + 'static;
 
     /// Attempts to create a new connection.
     async fn connect(&self) -> Result<Self::Connection, Self::Error>;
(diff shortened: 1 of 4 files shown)


marcbowes:
The tests will fail in the postgres/redis libraries because I haven't propagated the changes to make `Error` implement `Clone`. I thought we should discuss that first.

It seems like there are two paths forward:

1. If you get a catastrophic error, the Sink does _not_ get a copy of the actual error. Instead, it could get a `CatastrophicError` or something else. This seems reasonable to me.
2. Make `Error: Clone` as I've shown. This is not backwards compatible and is a fairly annoying bound to have.

marcbowes:
Thanks for the feedback.

> Yeah, I think it makes sense to not send the error to the sink if it's returned to the user directly. 

Agreed.

> I also think it makes sense to send something else, which should probably just contain a string containing a marked version of the Display representation of the original error.

This would require changing the signature of `ErrorSink` though. Currently it takes `Error` by value. Or is there something else you had in mind?

After thinking about this over lunch, I'm pretty comfortable just not sending it to the sink. What do you think?

> I don't think non-retryable is aptly represented as Catastrophic, though. Maybe just call it NonRetryable? (I think I prefer the y version over the i version, from some quick googling there doesn't seem to be a definitively right version either way.)

Sure thing.

djc:
Yes, I think your understanding of (1) is correct, and I do think it might solve your specific problem. I had forgotten about this exact feature, or I would have mentioned it sooner as potentially helpful to your problem. The initial spawning of connections can be executed asynchronously while returning the first error directly. Note, to make that work you'll need to configure `min_idle` to something other than 0 (the default).

djc:
Given the lack of follow-up, going to close this for now. Feel free to reopen if you think it's still relevant.

marcbowes:
My bad for losing track of this. It didn't solve my problem actually. I landed up just making an initial health check call myself, outside of bb8. I passed the client into bb8 so the initial (http) connection goes into the pool.

PR #107, opened 2021-07-12
Add `Pool::get_owned`

davidpdrsn:
Fixes https://github.com/djc/bb8/issues/72

This adds `Pool::get_owned` which returns `Result<PooledConnection<'static, M>, _>`. The `'static` means this pool is "detached" from the lifetime of the pool itself and can thus be used where the lifetime the pool gets in the way.

Considerations I made:

- Introduce a new connection type like `OwnedPooledConnection<M>` (similar to [`tokio::sync::OwnedSemaphorePermit`](https://docs.rs/tokio/1.8.1/tokio/sync/struct.OwnedSemaphorePermit.html)). However I imagine having two types of connections will be when you have to abstract over `OwnedPooledConnection` and `PooledConnection`.
- Make `PooledConnection` keep either a `std::sync::Weak<PoolInner<M>>` or a `&'a PoolInner<M>`. However that would require calling `Arc::upgrade` for most operations and how would one handle the case where that returns `None`? I guess pancking is the only option but I don't think thats ideal. I think storing an owned `PoolInner<M>` is nicer and leads to fewer surprises, though it does make it possible to leak the pool.

I have verified that this works for my use-case.

### DIFF
--- a/bb8/src/api.rs
+++ b/bb8/src/api.rs
@@ -55,14 +55,10 @@ impl<M: ManageConnection> Pool<M> {
         self.inner.get().await
     }
 
-    /// Retrieves an owned connection from the pool.
+    /// Retrieves an owned connection from the pool
     ///
-    /// [`Pool::get`] should be preferred for situations where an owned
-    /// connection type isn't necessary.
-    ///
-    /// This contains a reference to the pool and will prevent the it from being
-    /// dropped, while an owned connection is alive. So if an owned connection
-    /// is leaked, the pool will never be shutdown.
+    /// Using an owning `PooledConnection` makes it easier to leak the connection pool. Therefore, [`Pool::get`] 
+    /// (which stores a lifetime-bound reference to the pool) should be preferred whenever possible.
     pub async fn get_owned(&self) -> Result<PooledConnection<'static, M>, RunError<M::Error>> {
         self.inner.get_owned().await
     }
--- a/bb8/src/api.rs
+++ b/bb8/src/api.rs
@@ -365,14 +365,7 @@ where
     M: ManageConnection,
 {
     fn drop(&mut self) {
-        match &self.pool {
-            Cow::Borrowed(pool) => {
-                pool.put_back(self.conn.take());
-            }
-            Cow::Owned(pool) => {
-                pool.put_back(self.conn.take());
-            }
-        }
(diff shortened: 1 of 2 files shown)


bb8/src/inner.rs
```diff
@@ -83,13 +83,36 @@ where
...
+                inner: Arc::clone(&this.inner),
+            };
+            PooledConnection::new_owned(pool, conn)
+        })
+        .await
+    }
+
+    pub(crate) async fn get_with<'a, 'b, F>(
+        &'a self,
+        make_pool: F,
```
djc, line 104: Let's call this `make_pooled`.

bb8/src/api.rs
```diff
@@ -54,6 +55,18 @@ impl<M: ManageConnection> Pool<M> {
         self.inner.get().await
     }
 
+    /// Retrieves an owned connection from the pool.
```
djc, line 58:
```suggestion
    /// Retrieves an owned connection from the pool
```

bb8/src/api.rs
```diff
@@ -54,6 +55,18 @@ impl<M: ManageConnection> Pool<M> {
...
+    ///
+    /// [`Pool::get`] should be preferred for situations where an owned
+    /// connection type isn't necessary.
+    ///
+    /// This contains a reference to the pool and will prevent the it from being
+    /// dropped, while an owned connection is alive. So if an owned connection
+    /// is leaked, the pool will never be shutdown.
```
djc, line 65:
```suggestion
    /// Using an owning `PooledConnection` makes it easier to leak the connection pool. Therefore, [`Pool::get`] 
    /// (which stores a lifetime-bound reference to the pool) should be preferred whenever possible.
```

djc (commented):
This looks good modulo some minor nits. Thanks, and sorry for the long wait.

davidpdrsn:
Ready for another review!

bb8/src/inner.rs
djc: You used `make_pool` rather than `make_pooled`, which seems a little confusing. Mind fixing it?

davidpdrsn: whoops! Fixed

djc:
Thanks!

djc:
I've published bb8 0.7.1.

Issue #110, opened 2021-09-01
How to store and use values made during on_acquire?


cybermancy:
I have the following code:

```rust
#[derive(Debug)]
struct ConnectionCustomizer {}

#[async_trait]
impl bb8_postgres::bb8::CustomizeConnection<Client, bb8_postgres::tokio_postgres::Error> for ConnectionCustomizer {
    async fn on_acquire(
        &self,
        connection: & mut bb8_postgres::tokio_postgres::Client,
    ) -> Result<(), bb8_postgres::tokio_postgres::Error> {
        println!("test");
        Ok(())
    }
}
```

I want to prepare a number of queries using this for every connection. However, I'm struggling to think of a way to actually get the reference to the prepared statements for a specific connection. Is there a way to bind some data to the connection object, or at least a way to get some unique identifier for a connection that I could use to lookup their prepared statements in a hash table?

djc:
Maybe you should write your own `ManageConnection` instead of using bb8-postgres, which can deal with a type that is wrapped around `tokio_postgres::Client` (and potentially `Deref`s to it)?

djc:
(If it's generic enough I'd be happy to accept it into bb8-postgres and maintain it.)

Issue #57, opened 2020-01-24
Ergonomics issue with PostgresConnectionManager


ufoscout:
Hi, 

this issue is conceptually linked to https://github.com/sfackler/r2d2-postgres/issues/19

The problem is related to the fact that the `PostgresConnectionManager<T>` struct has a generic parameter `<T>`.
If `<T>` is known at compile-time, then there are no problems; on the contrary, if it is only known at runtime then it has to be redeclared and propagated in every place where the connection and/or the Pool is used.

As already shown in https://github.com/sfackler/r2d2-postgres/issues/19 , it is possible to slightly alter the `PostgresConnectionManager` struct implementation to remove the generic `T` and making the pool more ergonomic to use.

If you think it is not feasible, would you at least mind to provide both the alternatives of the `PostgresConnectionManager`, one with and one without the generic param?

khuey:
I wonder if this is really a bug with `tokio-postgres`. If it had an `impl MakeTlsConnect for Box<dyn MakeTlsConnect>`, then you could just box whatever you have into a trait object and things would work.

Any thoughts @sfackler?

t3hmrman:
Just want to pile on here to say I'm having the same issue -- in fact I'm trying to store the connection manager and the pool that gets created. At first I tried to use the code that @ufoscout wrote in sfackler/r2d2-postgres#19, but I didn't want to stray that far from the beaten path so I ended up with this monstrosity:

```rust
type ConnMgr = PostgresConnectionManager<Box<dyn MakeTlsConnect<Socket, TlsConnect=dyn Send, Stream=dyn Send, Error=dyn Error>>>;
```

Now I'm just running into the same problem with `Pool<M>` where `M` is something that implements `ManageConnection`, and I'm struggling on how to fill the `Connection` and `Error` type variables. While I'm sure this really cleaned up the code for `r2d2` it's made it a lot harder to use. Haven't been able to figure out how to do it for `Pool` so now I've moved making the pool to the area before I loop forever handling requests (it's a bit of an "actor"-ish setup).

leviinz:
I have nothing to add here, but I just wanted to comment that as a beginner to Rust, the usage of this is very confusing. Even as an experienced programmer just getting into Rust, it's extremely annoying to be able to accept any generic type for `PostgresConnectionManager`. From what I can tell, you can either use a type like `NoTls` directly (which I assume makes using TLS impossible?), or use the aforementioned type, which just looks ugly.

[...]

PR #112, opened 2021-10-24
Add example for wrapping PostgresConnectionManager

film42:
With this example we show how easy it is to wrap the postgres connection
pool to prepare several queries when a connection is created. Then, once
a connection is checked out, this examples shows how easy it is to use
the custom state to pull a prepared statement.

Inspired by #110 which requests this kind of behavior in bb8 directly,
but for now it's easy to extend any connection manager with custom
needs.

Happy hacktoberfest!

### DIFF
--- a/postgres/examples/custom_state.rs
+++ b/postgres/examples/custom_state.rs
@@ -1,0 +1,133 @@
+use std::collections::BTreeMap;
+use std::ops::Deref;
+use std::str::FromStr;
+
+use async_trait::async_trait;
+use bb8::{CustomizeConnection, Pool};
+use bb8_postgres::PostgresConnectionManager;
+use tokio_postgres::config::Config;
+use tokio_postgres::tls::{MakeTlsConnect, TlsConnect};
+use tokio_postgres::{Client, Error, Socket, Statement};
+
+// Select some static data from a Postgres DB
+//
+// The simplest way to start the db is using Docker:
+// docker run --name gotham-middleware-postgres -e POSTGRES_PASSWORD=mysecretpassword -p 5432:5432 -d postgres
+#[tokio::main]
+async fn main() {
+    let config =
+        tokio_postgres::config::Config::from_str("postgresql://postgres:docker@localhost:5432")
+            .unwrap();
+    let pg_mgr = CustomPostgresConnectionManager::new(config, tokio_postgres::NoTls);
+
+    let pool = Pool::builder()
+        .connection_customizer(Box::new(Customizer))
+        .build(pg_mgr)
+        .await
+        .expect("build error");
+
+    let connection = pool.get().await.expect("pool error");
+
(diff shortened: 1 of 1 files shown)


seanlinsley:
I'm trying to do something similar to this, but it doesn't seem like it works. For me at least, `pool.get()` returns a `PooledConnection` not a `CustomPostgresConnection` so there's no way to access the statement cache.

djc:
I recommend you start from this exact example, which is compile checked in CI and thus should work.

seanlinsley:
Ah, it looks like the connection having a type of `PooledConnection` is expected. I get the same error when trying to specify the type in the example code:

```rust
let connection: CustomPostgresConnection = pool.get().await.expect("pool error");
```

So my issue seems to be more related to #57. It's not clear what type annotations are needed so bb8 pooled connections can be passed to a function. This for example fails with:
>no method named `query` found for struct `PooledConnection<'_, M>` in the current scope
```rust
async fn run_query<M: bb8::ManageConnection>(connection: PooledConnection<'_, M>) {
    connection.query("", &[]).await.unwrap();
}
```
Sorry if this has been asked somewhere else, but if it has I can't find it.

djc:
Your code sample presupposes the existence of a `query()` method on `M::Connection`. You'll need to explain to the compiler why it does. For example, `M` is some specific type, like `CustomPostgresConnectionManager`, or `M::Connection` implements some trait that provides a `query()` method of the appropriate type.

seanlinsley:
Right, so how do I do that? 😅 The [postgres glue code](https://github.com/djc/bb8/blob/main/postgres/src/lib.rs) isn't implementing a trait from the tokio_postgres crate so it's not clear how these function calls are passed through to the postgres connection as it is.

djc:
Make it abstract over `dyn MakeTlsConnect<Socket>`?

seanlinsley:
`dyn MakeTlsConnect<Socket>` gets me to:

>help: specify the associated types: `MakeTlsConnect<Socket, Stream = Type, TlsConnect = Type, Error = Type>`

I tried multiple versions of that but couldn't work around this error:

>`(dyn MakeTlsConnect<...> + 'static)` cannot be shared between threads safely

I also tried defining a type like https://github.com/djc/bb8/issues/57#issuecomment-613304110 but ran into the same error.

djc:
Add a `Send` bound. But all of this ends up being Rust knowledge/experience, so maybe better on Stack Overflow or users.rust-lang.org.

seanlinsley:
Yes I don't have much experience with Rust, but this is a basic usability / documentation issue with bb8. I've tried multiple variations here, including adding `Send`, and I'm thrown into an ever-deeper hole of errors.

Re: posting this elsewhere, I doubt anyone's going to put in the time to meaningfully help if they're not already familiar with this project.

Issue #95, opened 2021-02-03
Task mixing


invis87:
Copypasted of issue that I create in tokio: https://github.com/tokio-rs/tokio/issues/3500

**Version**
```
    │   ├── tokio v1.1.0
    │   │   └── tokio-macros v1.0.0
    │   ├── tokio-util v0.6.2
    │   │   ├── tokio v1.1.0 (*)
    │   │   └── tokio-stream v0.1.2
    │   │       └── tokio v1.1.0 (*)
    │   ├── tokio v1.1.0 (*)
    ├── tokio v1.1.0 (*)
    ├── tokio-stream v0.1.2 (*)
    ├── tokio-util v0.6.2 (*)
    │   ├── tokio v1.1.0 (*)
    │   ├── tokio-stream v0.1.2 (*)
│   └── tokio v1.1.0 (*)
│       │   └── tokio v1.1.0 (*)
│       ├── tokio v1.1.0 (*)
│       ├── tokio-util v0.6.2 (*)
├── tokio v1.1.0 (*)
├── tokio-util v0.6.2 (*)
```

`cargo tree | grep redis`:
```
├── bb8-redis v0.8.0
│   └── redis v0.19.0
```

`cargo tree | grep bb8`:
```
├── bb8 v0.7.0
├── bb8-redis v0.8.0
│   ├── bb8 v0.7.0 (*)
```

**Platform**
`Linux videotest01.dev.wb.ru 4.19.0-13-cloud-amd64 #1 SMP Debian 4.19.160-2 (2020-11-28) x86_64 GNU/Linux`

**Description**
After moving to Tokio v1.0.1 our "reddis-proxy" service start to fail with "magic" errors:
`Response was of incompatible type: "Response type not hashmap compatible" (response was status("PONG"))`
This error happens when we call `let response: redis::RedisResult<collections::HashMap<String, String>> =        conn.hgetall(key).await;`
Looks like async tasks mixed somehow and `hgetall` get response from `PING` request. Do not know where to dig into, will try with newer version of tokio...

--------------------------------------

We have async fn:
```
async fn ask_redis(
    redis_pool: Pool<RedisConnectionManager>,
) -> Result<_> {
    let mut conn = redis_pool.get().await?;
    let key = todo!();

    let response: redis::RedisResult<collections::HashMap<String, String>> =
        conn.hgetall(key).await;
    
    ...
}
```
[...]

PR #114, opened 2021-11-10
Implement `multiplexed-connections` feature

Barre:
Implement multiplexed connections in order to implement a workaround against issue #95.
This workaround was mentioned by the author of the redis-rs library: https://github.com/mitsuhiko/redis-rs/issues/325#issuecomment-846007813

Based on my experience of the last few weeks, the issue didn't occur anymore after using `MultiplexedConnection` instead of `Connection`.

### DIFF
--- a/redis/Cargo.toml
+++ b/redis/Cargo.toml
@@ -6,10 +6,15 @@ license = "MIT"
 repository = "https://github.com/djc/bb8"
 edition = "2018"
 
+[features]
+multiplexed-connections = []
+
 [dependencies]
 async-trait = "0.1"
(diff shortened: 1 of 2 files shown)


djc:
Hey, thanks. In the issue I asked if you could explain a bit more about why/how the semantics are different.

I don't think doing this as a feature that changes the connection type makes sense. It could just be a separate `ManageConnection` implementation in bb8-redis, or potentially we should just semver-bump bb8-redis and always use the `MultiplexedConnection`, since you'll pretty much want these semantics for pooling anyway.

PR #116, opened 2021-11-19
Restore ManageConnection::is_valid original signature

arnodb:
This addresses https://github.com/djc/bb8/issues/115.

I haven't been in position to test it unfortunately, but it compiles.


djc:
Thanks!

PR #124, opened 2022-05-02
Add reap_all_idle_connections option, fixes #123

### LINKED ISSUE #123: Reaper closes connections in the pool even when it violates min_idle
According to the docs for [`Builder::idle_timeout`](https://docs.rs/bb8/latest/bb8/struct.Builder.html#method.idle_timeout), "idle connections in excess of `min_idle` will be closed at the next reaping after remaining idle past this duration." To me, this implies that connections that are not in excess of `min_idle` are kept alive; in my opinion, this is desirable, since creating a connection could be expensive.

However, it appears that the reaper closes any connections that have been alive or idle for too long; based on `max_lifetime` and `idle_timeout`, respectively. It then respawns enough connections to have `min_idle` connections available. This seems to be inconsistent with the docs, and is also less desirable in my opinion. My use case is:

- Creating connections is expensive
- I would like to keep one connection open at all times (`min_idle == 1`)
- I would like to close extra connections relatively quickly (for instance, `idle_timeout == Duration::from_secs(1)`)

However, this means that even if my application never requests a connection, the idle connection is refreshed basically every time the reaper is scheduled.

connorjayr:
Adds a new boolean option to Builder named `reap_all_idle_connections`. If true (the default value), the current behavior of the pool is maintained for backward compatibility. However, if false, the reaper will not close any idle connections if it results in there being fewer than `min_idle` connections, even if the connections have been idle for longer than `idle_timeout`.

This option would best be disabled if connections are expensive to establish and/or are generally stable over time.

### DIFF
--- a/bb8/src/api.rs
+++ b/bb8/src/api.rs
@@ -87,6 +87,9 @@ pub struct Builder<M: ManageConnection> {
     pub(crate) max_lifetime: Option<Duration>,
     /// The duration, if any, after which idle_connections in excess of `min_idle` are closed.
     pub(crate) idle_timeout: Option<Duration>,
+    /// Whether or not to reap all connections idle for longer than
+    /// `idle_timeout`, rather than just those in excess of `min_idle`.
+    pub(crate) reap_all_idle_connections: bool,
     /// The duration to wait to start a connection before giving up.
     pub(crate) connection_timeout: Duration,
     /// The error sink.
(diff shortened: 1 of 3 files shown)


djc:
So to clarify, your beef with the current implementation is that it reaps connections that have been idle longer than the `idle_timeout` even if doing so drops the number of idle connections below `min_idle`?

I think that's reasonable, but I think for now we shouldn't introduce yet another option for it, just making it the default. I'm assuming that getting to `max_lifetime`, unlike `idle_timeout` *should* probably still override the `min_idle` boundary, do you agree?

connorjayr:
That was my initial line of thought, yeah. Although after putting more consideration into it, I can imagine that there is a valid use case for wanting connections to recycle after being idle for too long (e.g. if connections are dropped by the server after being idle for too long), and I would be worried that changing the default behavior of the library to no longer do this could break some users' applications.

As for my personal use case, I realized that making the idle timeout a bit longer would be fine. Holding (a small number of) connections open is not the end of the world, and so an idle timeout of 10 minutes or so--which is the default, IIRC--works for me. Still, it might be nice to have an option to allow idle connections to live forever, since connections are typically validated on check out anyways.

PR #125, opened 2022-06-07
Bump to 2021 edition, inline formats, lints

nyurik:
* Bump Rust edition 2018 -> 2021
* Use new inline formatting, i.e. `format!("{v}")` instead of `format!("{}", v)` (simple cases only)
* use more specific asserts like `assert_ne!`
* ignore intellij IDE files
* remove a few un-needed namespace prefixes


djc:
Hey, thanks for this. This is a bit of a grab bag of changes. Are you gearing up for further contributions here, what's the angle?

Mostly it looks good. I think I'd like to stick with 2018 edition for now so we can keep a lower minimum supported Rust version for now. I don't think we test that explicitly in CI yet, maybe you could add that, and see what's the oldest version that currently works?

djc:
Yes, we should stick to less than 1.56, would be nice if you can set that in CI and add a note to the README.

nyurik:
I moved the simpler ones to a separate PR.  CI changes are still a TODO

djc:
Since we're on hashbrown 0.12 now which requires Rust 1.56 anyway, going to merge this. Thanks!

djc (approved):

Issue #131, opened 2022-07-30
Considerable number of pedantic lints reported by clippy


nyurik:
Hi, I ran `cargo clippy -- -W clippy::pedantic --no-deps` and it seems there are quiet a few lints being reported, some possibly more important than others.  Are you interesting in a fix for any of those, or do you think they are ok to ignore? Thx!

<details>

<summary>full output</summary>

```
warning: redundant else block
   --> bb8/src/inner.rs:219:28
    |
219 |                       } else {
    |  ____________________________^
220 | |                         delay = max(Duration::from_millis(200), delay);
221 | |                         delay = min(self.inner.statics.connection_timeout / 2, delay * 2);
222 | |                         sleep(delay).await;
223 | |                     }
    | |_____________________^
    |
    = note: `-W clippy::redundant-else` implied by `-W clippy::pedantic`
    = help: remove the `else` block and move the contents out
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_else

warning: this method could have a `#[must_use]` attribute
  --> bb8/src/api.rs:44:5
   |
44 |     pub fn builder() -> Builder<M> {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: add the attribute: `#[must_use] pub fn builder() -> Builder<M>`
   |
   = note: `-W clippy::must-use-candidate` implied by `-W clippy::pedantic`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#must_use_candidate

warning: this method could have a `#[must_use]` attribute
  --> bb8/src/api.rs:49:5
   |
49 |     pub fn state(&self) -> State {
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: add the attribute: `#[must_use] pub fn state(&self) -> State`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#must_use_candidate

warning: docs for function returning `Result` missing `# Errors` section
  --> bb8/src/api.rs:54:5
   |
54 | /     pub async fn get(&self) -> Result<PooledConnection<'_, M>, RunError<M::Error>> {
55 | |         self.inner.get().await
56 | |     }
   | |_____^
   |
   = note: `-W clippy::missing-errors-doc` implied by `-W clippy::pedantic`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#missing_errors_doc

[...]

PR #132, opened 2022-07-30
Some minor lints

nyurik:
* use `#[must_use]` for Self-returning fn
* some minor docs for panics
* use `Self` instead of a specific type in some cases

I used this command to test:

```
❯ cargo clippy -- -W clippy::default_trait_access -W clippy::missing_panics_doc -W clippy::return-self-not-must-use -W clippy::semicolon-if-nothing-returned -W clippy::trait-duplication-in-bounds
```

There is only one issue remaining that I'm not certain how to fix (or if its even needed to be fixed?):

```
warning: this trait bound is already specified in the where clause
  --> postgres/src/lib.rs:52:46
   |
52 |     <Tls as MakeTlsConnect<Socket>>::Stream: Send + Sync,
   |                                              ^^^^
   |
   = note: requested on the command line with `-W clippy::trait-duplication-in-bounds`
   = help: consider removing this trait bound
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#trait_duplication_in_bounds
```

See #131 .  I tried to add a clippy.toml file to make this more consistent, but that seems not to work properly -- https://github.com/rust-lang/rust-clippy/issues/9265

### DIFF
--- a/bb8/src/api.rs
+++ b/bb8/src/api.rs
@@ -119,14 +119,20 @@ impl<M: ManageConnection> Builder<M> {
     /// Constructs a new `Builder`.
     ///
     /// Parameters are initialized with their default values.
-    pub fn new() -> Builder<M> {
-        Default::default()
+    #[must_use]
+    pub fn new() -> Self {
+        Builder::default()
     }
 
     /// Sets the maximum number of connections managed by the pool.
     ///
     /// Defaults to 10.
-    pub fn max_size(mut self, max_size: u32) -> Builder<M> {
+    ///
+    /// # Panics
+    ///
+    /// Will panic if `max_size` is 0.
+    #[must_use]
+    pub fn max_size(mut self, max_size: u32) -> Self {
         assert!(max_size > 0, "max_size must be greater than zero!");
         self.max_size = max_size;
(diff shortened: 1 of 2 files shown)


bb8/src/api.rs
```diff
@@ -119,14 +119,18 @@ impl<M: ManageConnection> Builder<M> {
...
+        Builder::default()
     }
 
     /// Sets the maximum number of connections managed by the pool.
     ///
     /// Defaults to 10.
-    pub fn max_size(mut self, max_size: u32) -> Builder<M> {
+    ///
+    /// # Panics
+    /// Will panic if `max_size` is 0
```
djc, line 131: Style nit: can you add an empty line after `# Panics` and a period at the end of the following sentence, here and for the other affected cases?

djc (approved):
Thanks, this is looking pretty good!

djc:
Looks to me like the trait-duplication-in-bounds issue is a clippy false positive, might be worth filing an issue about that.

nyurik:
thx, fixed

djc (approved):

PR #134, opened 2022-08-01
Set MSRV to 1.58 for now

djc:


djc:
@nyurik do you know what might be going on with the CI failure here?

nyurik:
@djc the MSRV must be set to 1.58, not 1.56.  That's the first version with inline format strings. See https://github.com/djc/bb8/pull/125#issuecomment-1149072876

PR #135, opened 2022-08-31
Fix readme example

crsov:
This alllows to simply replace `bb8_foodb::FooConnectionManager::new("localhost:1234")` with `bb8_redis::RedisConnectionManager::new("redis://localhost:1234").unwrap()` as an example and code will compile!


djc (approved):
Thanks!

PR #144, opened 2022-12-02
Add support for customized future executor

gaoqiangz:
Support to customize the runtime associated with `spawn` to solve the problem of across runtimes.
Scenario:
The connection connected by the runtime `A`, use by the runtime `B`, destroy `A` before `B` while connection are using will cause `IO driver has terminated`.

### DIFF
--- a/bb8/src/api.rs
+++ b/bb8/src/api.rs
@@ -1,15 +1,21 @@
-use std::borrow::Cow;
-use std::error;
-use std::fmt;
-use std::marker::PhantomData;
-use std::ops::{Deref, DerefMut};
-use std::time::Duration;
-
-use async_trait::async_trait;
-
 use crate::inner::PoolInner;
 use crate::internals::Conn;
 pub use crate::internals::State;
+use async_trait::async_trait;
+use std::borrow::Cow;
+use std::error;
+use std::fmt;
+use std::future::Future;
+use std::marker::PhantomData;
+use std::ops::{Deref, DerefMut};
+use std::pin::Pin;
+use std::time::Duration;
+
+/// An executor of futures.
+pub trait Executor: Sync + Send + 'static {
+    /// Place the future into the executor to be run.
+    fn execute(&self, fut: Pin<Box<dyn Future<Output = ()> + Send>>);
+}
 
 /// A generic connection pool.
 pub struct Pool<M>
@@ -253,7 +259,7 @@ impl<M: ManageConnection> Builder<M> {
         self
     }
 
-    fn build_inner(self, manager: M) -> Pool<M> {
+    fn build_inner<E: Executor>(self, manager: M, executor: E) -> Pool<M> {
         if let Some(min_idle) = self.min_idle {
             assert!(
                 self.max_size >= min_idle,
@@ -262,7 +268,7 @@ impl<M: ManageConnection> Builder<M> {
         }
 
         Pool {
-            inner: PoolInner::new(self, manager),
+            inner: PoolInner::new(self, manager, executor),
         }
     }
 
@@ -271,7 +277,19 @@ impl<M: ManageConnection> Builder<M> {
     /// The `Pool` will not be returned until it has established its configured
     /// minimum number of connections, or it times out.
     pub async fn build(self, manager: M) -> Result<Pool<M>, M::Error> {
-        let pool = self.build_inner(manager);
+        self.build_with_executor(manager, TokioExecutor).await
+    }
+
+    /// Consumes the builder with the specified executor, returning a new, initialized `Pool`.
+    ///
+    /// The `Pool` will not be returned until it has established its configured
+    /// minimum number of connections, or it times out.
+    pub async fn build_with_executor<E: Executor>(
+        self,
+        manager: M,
+        executor: E,
+    ) -> Result<Pool<M>, M::Error> {
+        let pool = self.build_inner(manager, executor);
         pool.inner.start_connections().await.map(|()| pool)
     }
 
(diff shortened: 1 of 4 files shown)


djc:
What runtime are you using? Why do you want to use bb8 instead of another connection pool that already supports your runtime?

gaoqiangz:
Please see this example:

- Cargo.toml

```toml
bb8 = "0.8.0"
bb8-tiberius = { version = "0.13.0", features = ["sql-browser"] }
```

- main.rs

```rust
#[tokio::main]
async fn main() -> anyhow::Result<()> {
    let host = ".\\JQZ"; //local instance `JQZ`
    let conn_str = format!(
        "Server={host};Database=master;TrustServerCertificate=true;IntegratedSecurity=true;"
    );

    let mgr = bb8_tiberius::ConnectionManager::build(conn_str.as_str())?.using_named_connection();
    let pool = bb8::Pool::builder().max_size(1).build_unchecked(mgr);

    let (tx_connected, rx_connected) = tokio::sync::oneshot::channel();
    let thread_1 = std::thread::spawn({
        let pool = pool.clone();
        move || {
            let rt = tokio::runtime::Builder::new_current_thread()
                .enable_all()
                .build()
                .unwrap();
            rt.block_on(async {
                //connect in this runtime
                println!("thread_1 connecting");
                println!("thread_1 connected");
                tx_connected.send(());
                //make `thread_2` block at the `get` call
                tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
                //release `thread_2`
                println!("thread_1 release");
                drop(conn);
                tokio::time::sleep(tokio::time::Duration::from_secs(0)).await;
            });
            println!("thread_1 runtime dropping");
            //cause `thread_2` error
            drop(rt);
            println!("thread_1 runtime droped");
        }
    });

    let thread_2 = std::thread::spawn({
        let pool = pool.clone();
        move || {
            let rt = tokio::runtime::Builder::new_current_thread()
                .enable_all()
                .build()
                .unwrap();
            rt.block_on(async {
                rx_connected.await;
                //use here
                println!("thread_2 getting connection");
                let mut conn = pool.get().await.unwrap();
                println!("thread_2 querying");
                //sleep 10 seconds, should have a error `IO driver has terminated`
                let e = conn.simple_query("WAITFOR DELAY '00:00:10'").await.err();
                println!("thread_2 queried: {e:?}");
            });
        }
    });

    thread_1.join().unwrap();
    println!("thread_1 exited");
    thread_2.join().unwrap();
    println!("thread_2 exited");

    Ok(())
}
```

- log

```
thread_1 connecting
thread_1 connected
thread_2 getting connection
thread_1 release
thread_2 querying
thread_1 runtime dropping
thread_2 queried: Some(Io { kind: Other, message: "IO driver has terminated" })
thread_1 runtime droped
thread_1 exited
thread_2 exited
```

djc:
Okay, so you want to use a per-thread single-threaded Tokio runtime. Could have just said so...

gaoqiangz:
Yes, there are multiple runtimes in our program,such as independent timer threads.

bb8/src/inner.rs
```diff
@@ -1,18 +1,15 @@
+use crate::api::{Builder, Executor, ManageConnection, PooledConnection, RunError};
+use crate::internals::{Approval, ApprovalIter, Conn, SharedPool, State};
+use futures_channel::oneshot;
+use futures_util::stream::{FuturesUnordered, StreamExt};
```
djc, line 4: Please undo this unrelated change. I would like to keep the imports in three separate blocks: `std`, external crates and crate-local (in that order).

bb8/src/internals.rs
```diff
@@ -1,12 +1,10 @@
-use std::cmp::min;
-use std::sync::Arc;
-use std::time::Instant;
-
+use crate::api::{Builder, Executor, ManageConnection};
 use futures_channel::oneshot;
 use parking_lot::Mutex;
-
-use crate::api::{Builder, ManageConnection};
+use std::cmp::min;
```
djc, line 4: Please undo this unrelated change.

bb8/src/api.rs
```diff
@@ -1,15 +1,21 @@
...
 use std::ops::{Deref, DerefMut};
+use std::pin::Pin;
 use std::time::Duration;
 
-use async_trait::async_trait;
-
-use crate::inner::PoolInner;
-use crate::internals::Conn;
-pub use crate::internals::State;
+/// An executor of futures.
```
djc, line 14: I'd like this definition to go at the bottom of the module, with the `struct TokioExecutor` definition directly above it.

bb8/src/api.rs
```diff
@@ -1,15 +1,21 @@
...
+pub trait Executor: Sync + Send + 'static {
+    /// Place the future into the executor to be run.
+    fn execute(&self, fut: Pin<Box<dyn Future<Output = ()> + Send>>);
```
djc, line 17: Why does this need to take a `Pin<Box<_>>` instead of just an `impl Future`?

bb8/src/inner.rs
```diff
@@ -254,11 +251,14 @@ where
     }
 }
 
-fn schedule_reaping<M>(mut interval: Interval, weak_shared: Weak<SharedPool<M>>)
-where
+fn schedule_reaping<M>(
+    executor: &Box<dyn Executor>,
```
djc: Please address the lint issue here.

bb8/src/api.rs
gaoqiangz: `Executor` should be object-safe trait:
https://github.com/gaoqiangz/bb8/blob/efef42610d1a8a4d546513a621cb5af98214e7bf/bb8/src/internals.rs#L18

bb8/Cargo.toml
```diff
@@ -1,9 +1,9 @@
 [package]
 name = "bb8"
-version = "0.8.0"
-description = "Full-featured async (tokio-based) connection pool (like r2d2)"
+version = "0.8.0-rc.1"
```
djc, line 3: I won't be able to merge this PR with this stuff in it, please revert these changes.

bb8/src/api.rs
```diff
@@ -456,3 +478,17 @@ impl<E> ErrorSink<E> for NopErrorSink {
...
+}
+
+struct TokioExecutor;
+
+impl Executor for TokioExecutor {
+    fn execute(&self, fut: Pin<Box<dyn Future<Output = ()> + Send>>) {
```
djc, line 491: Would prefer to call this `spawn()`.

bb8/src/internals.rs
```diff
@@ -15,6 +15,7 @@ where
     M: ManageConnection + Send,
 {
     pub(crate) statics: Builder<M>,
+    pub(crate) executor: Box<dyn Executor>,
```
djc, line 18: Why does this need a trait object? It would be nice if we didn't have to `Box` every future just to support this fairly niche use case.

PR #147, opened 2023-01-05
fix #141

### LINKED ISSUE #141: Connection pool hangs if authentication fails
When using a connection pool with tokio_postgres, the connection hangs until a TimedOut error is returned if the password is incorrect. This behaviour is undesirable and should just result in authentication failure.

To re-produce, simply try connect to any database using an incorrect password and the connection will hang until it times out.

goldwind-ting:
Is it reasonable to check the connection in `build`? @djc

### DIFF
--- a/bb8/src/api.rs
+++ b/bb8/src/api.rs
@@ -272,6 +272,7 @@ impl<M: ManageConnection> Builder<M> {
     /// minimum number of connections, or it times out.
     pub async fn build(self, manager: M) -> Result<Pool<M>, M::Error> {
         let pool = self.build_inner(manager);
+        pool.inner.preconnect().await?;
         pool.inner.start_connections().await.map(|()| pool)
     }
 
--- a/bb8/src/inner.rs
+++ b/bb8/src/inner.rs
@@ -152,7 +152,7 @@ where
     }
 
     pub(crate) async fn preconnect(&self) -> Result<(), M::Error> {
-        self.inner.manager.connect().await.and_then(|_|Ok(()))
+        self.inner.manager.connect().await.map(|_| ())
     }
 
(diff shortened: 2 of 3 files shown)


djc:
This seems more like a workaround than a fix for the issue described in #141 -- AFAICT in this case the connection would still time out, but it would happen at pool instantiation time rather than at `pool.get()` time?

goldwind-ting:
> This seems more like a workaround than a fix for the issue described in #141 -- AFAICT in this case the connection would still time out, but it would happen at pool instantiation time rather than at `pool.get()` time?

Yes, it would still time out, but I think in this case it is caused possibly by network.
If merged the PR, it will return directly when failed to check configure, e.g. password, database name.
And if you think this is not reasonable, I will try another solution, you can help me review.

djc:
Yes, I'd like to understand why we're timing out in that case instead of immediately-ish getting an error about the password being wrong.

goldwind-ting:
> Yes, I don't think this is a good solution. I'd like to understand why we're timing out in that case instead of immediately-ish getting an error about the password being wrong.

Seem that when failed to connect database, it would `loop` until [timeout](https://github.com/djc/bb8/blob/656b72da0d8e4dd9cea75994efff682997a29c7f/bb8/src/inner.rs#L214-L223) and do not send `guard` to `waiters` [channel](https://github.com/djc/bb8/blob/656b72da0d8e4dd9cea75994efff682997a29c7f/bb8/src/inner.rs#L142-L145).

djc:
Right, so I think the correct fix would be to check the error returned by `connect()` in that case and avoid looping if we can determine the problem is likely to be permanent.

Issue #102, opened 2021-03-25
Handling invalid connections


marcbowes:
If a user incorrectly specifies their connection setting (e.g. typos the IP address), then bb8's `get` method will hit errors and retry until the timeout. Is there a way of marking certain errors as "just give up"?

djc:
There is currently no such way that I'm aware. The question is if this is really what you want -- it seems likely that many of these kinds of errors could be transient, so retrying them before timing out might actually be desirable.

marcbowes:
Yeah, I agree with you in general. In my case, we have a shell utility connecting to a db. If the user typos their db name then it takes 30s (the timeout) to return an error. I know logging with the error sink will help, but really, retrying is pointless.

djc:
Maybe you should just set the timeout lower in your application?

marcbowes:
Well, that’s not the only problem. The other is that the error message just says timeout. 

Do you think bb8 should not have the concept of an exception that should not be retried on (is terminal for the get attempt and is exposed to the user)?

djc:
That's a pretty abstract question, to which I don't have a concrete answer. I'm skeptical that there's a straightforward way to implement this that doesn't make things worse for other users and doesn't incur a whole bunch of complexity for what in my opinion is limited benefit.

bb8 is chiefly intended for long-running (server) processes. Optimizing for startup doesn't necessarily make sense. You might just start a connection without bb8 to start with to validate the user's input.

marcbowes:
The simplest thing I could think of would be to add 
```rust
fn should_retry(err: &Self::Error) -> bool
``` 

to `ManageCollection` with a default impl that returns true. That's backwards compatible and doesn't "incur a whole bunch of complexity" in my opinion.

The `get` API already supports exposing the manager defined Error type:

```
pub async fn get(&self) -> Result<PooledConnection<'_, M>, RunError<M::Error>> 
```

I'm probably missing something, though.

djc:
That seems fine to me, let's have a PR (please add some tests).

marcbowes:
I ran into an issue where `ManagedConnection::Error` is not `Clone`. The problem arises because both `get` and `ErrorSink` want the error by value, so there needs to be a copy made. Currently, only the sink gets it because `get` always returns `TimedOut` and never `UserError(M::Error)`. Not sure how to address this without adding a new bound which would not be backwards compatible.

marcbowes:
I opened https://github.com/djc/bb8/pull/104 to discuss the path forward.

PR #149, opened 2023-01-25
Add option to retry connection creation, on by default

levkk:
Fix for #102

Overall I agree with the point that connection retries are good and transient errors are common. However, sometimes the application knows better. E.g. we use bb8 in https://github.com/levkk/pgcat and during database restarts, we don't want bb8 to retry because we know the database is down (just for a few seconds), and we have logic to handle and back-off in our code. In our situation, we have thousands of clients creating a thundering herd of retries.

This PR makes the retries optional, enabled by default to preserve existing behavior.

### DIFF
--- a/bb8/src/api.rs
+++ b/bb8/src/api.rs
@@ -228,7 +228,7 @@ impl<M: ManageConnection> Builder<M> {
         self
     }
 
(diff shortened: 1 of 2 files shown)


djc:
Thanks!

PR #152, opened 2023-02-09
bb8: Make parking_lot an optional feature enabled by default

r-ml:
Rust 1.62 improved `std::sync::Mutex` speed and 1.63 `const`ified its constructor. It can now be used in the same contexts `parking_lot`'s implementation can. 

Currently, `bb8` is enabling the `parking_lot` tokio feature for its users. A library should probably not do that without a good reason. With the recent-ish improvements to `std::sync::Mutex`, I believe we can drop `parking_lot` from the dependency tree entirely.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -45,7 +45,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: 1.58.0
+          toolchain: 1.60.0
           override: true
       - uses: actions-rs/cargo@v1
         with:
--- a/bb8/Cargo.toml
+++ b/bb8/Cargo.toml
@@ -13,8 +13,12 @@ rust-version = "1.60"
 async-trait = "0.1"
 futures-channel = "0.3.2"
 futures-util = { version = "0.3.2", default-features = false, features = ["channel"] }
-parking_lot = "0.12"
-tokio = { version = "1.0", features = ["rt", "time", "parking_lot"] }
+parking_lot = { version = "0.12", optional = true }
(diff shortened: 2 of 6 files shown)


djc:
Well, bb8 currently has an MSRV of 1.57, so we cannot rely on the `const` constructor. On top of that, the 1.62 release notes only discuss Linux, not any of the other platforms. So I don't think this makes sense as a default. I'd be okay to take this as a Cargo feature that is enabled by default, though.

r-ml:
Done. Please tell me whether the `mod` is a good enough way to implement this or would you prefer I sprinkle `#[cfg(feature ...)]` everywhere.

r-ml:
Well, cargo's `dep:` feature was added in Rust 1.60 *sigh*...

r-ml:
I installed 1.57 and I can work around the issue by naming the feature `parking-lot` instead of `parking_lot`. Let me know whether that's good enough.

Another thing, compilation failed [here](https://github.com/djc/bb8/blob/main/bb8/src/api.rs#L411) with "there is no argument named `err`". This appears to be fixed with 1.58. Want me to make a PR to make it compile? CI should have caught that though :thinking:

djc:
Okay, let's bump the MSRV to 1.60 (in a separate commit) so we can use `dep:`?

The CI setup, as I recall it, is a bit strange because the redis crate has a newer MSRV than the core bb8 crates. But bumping the MSRV should fix this anyway.

r-ml:
Done. (For some reason I did not get notified by Github of your comment)

djc (approved):
Thanks!

PR #153, opened 2023-03-20
Forward error to waiter before sinking

couchand:
This ensures that an active waiter will be notified of a connection failure before we resort to sending it to the sink.

Contributes to #141.


couchand:
I would observe that, although I think this generally makes sense, since it doesn't fully fix #141 and dealing with the `Result` channel is a bit awkward in places I'm not 100% sold on this PR as-is.  Happy to hear any feedback.

djc:
This seems like a reasonable improvement to me.

PR #158, opened 2023-05-03
Set reaper rate to minimum value

### LINKED ISSUE #157: Reaper default frequency conflicts with other settings
The reaper runs every 30 seconds by default, which means any `max_lifetime` and `idle_timeout` will be ignored (well, enforced too late really) if set to less than 30 seconds.

A good automatic default could be to automatically set `reaper_rate` to `min(max_lifetime, idle_timeout, 30)`.

levkk:
Fix #157

### DIFF
--- a/bb8/src/api.rs
+++ b/bb8/src/api.rs
@@ -189,8 +189,6 @@ impl<M: ManageConnection> Builder<M> {
 
         if let Some(max_lifetime) = max_lifetime {
             self.reaper_rate = std::cmp::min(DEFAULT_REAPER_RATE, max_lifetime);
-        } else {
-            self.reaper_rate = std::cmp::min(DEFAULT_REAPER_RATE, self.reaper_rate)
         }
 
         self.max_lifetime = max_lifetime;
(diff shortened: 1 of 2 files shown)


bb8/src/api.rs
```diff
@@ -183,6 +186,11 @@ impl<M: ManageConnection> Builder<M> {
             Some(Duration::from_secs(0)),
             "max_lifetime must be greater than zero!"
         );
+
+        if let Some(max_lifetime) = max_lifetime {
```
djc, line 190: This isn't very precise... if `max_lifetime` is 30s and the reaper rate is set to 30s as well, if the reaper comes along after 15s it won't reap and then it will be 45s before the `max_lifetime` is enforced. I'm inclined to think this is more like a documentation problem, and/or maybe we could emit some warnings if there is a large difference between these values?

levkk: True. Alternatively, we can just run the reaper more frequently, e.g. [PgBouncer](https://github.com/pgbouncer/pgbouncer) runs its "janitor" 3 times per second.

djc: That's what the configuration is for. I'm comfortable with the current default configuration as a default, so the goal here is to try and make sure people don't get surprised by the interactions.

levkk: Well... if we look at how the interface is, right now someone can set `max_lifetime` to `1s` and nothing will happen for 30 seconds. So ideally, someone sets that setting to any value they want and the pool just works. `reaper_rate` is a leaked abstraction imo, it requires the user to understand how the pool works internally.

djc: I don't disagree that it's a leaky abstraction, my point is more that I don't think it's feasible to avoid the leakage in a fairly general-purpose abstraction like this. So instead of trying do so in a limited way which could be surprising in different ways, I'd prefer clearly documenting how these values work together.

PR #159, opened 2023-05-15
add pool for pubsub connections

bbaldino:
Based on discussion in https://github.com/djc/bb8/issues/70

### DIFF
--- a/redis/src/lib.rs
+++ b/redis/src/lib.rs
@@ -39,8 +39,49 @@ pub use bb8;
 pub use redis;
 
 use async_trait::async_trait;
+use redis::aio::PubSub;
 use redis::{aio::Connection, ErrorKind};
 use redis::{Client, IntoConnectionInfo, RedisError};
+
+/// A `bb8::ManageConnection` for `redis::aio::PubSub`
+#[derive(Clone, Debug)]
+pub struct RedisPubSubConnectionManager {
+    client: Client,
+}
+
+impl RedisPubSubConnectionManager {
+    /// Create a new `RedisConnectionPubSubManager`.
+    /// See `redis::Client::open` for a description of the parameter types.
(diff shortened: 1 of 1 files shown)


redis/src/lib.rs
```diff
@@ -39,9 +39,50 @@ pub use bb8;
...
+impl bb8::ManageConnection for RedisPubSubConnectionManager {
+    type Connection = PubSub;
+    type Error = RedisError;
+
+    async fn connect(&self) -> Result<Self::Connection, Self::Error> {
+        Ok(self.client.get_async_connection().await?.into_pubsub())
+    }
+
+    async fn is_valid(&self, _conn: &mut Self::Connection) -> Result<(), Self::Error> {
+        // TODO:
```
bbaldino, line 72: Something needs to be done here, but not sure what would be best for a PubSub.

djc: Maybe `punsubscribe()` with an empty pattern?

bbaldino: Good idea.  Made that change, just trying to validate a few things.

bbaldino:
Ok, pushed the logic for `is_valid`.  Unfortunately I wasn't able to really validate some things in my use case (due to other issues), and didn't see any unit tests in the redis that I could expand on.   If there's anything you think might be risky here (bug-wise) we can hold off.

djc:
> Ok, pushed the logic for `is_valid`. Unfortunately I wasn't able to really validate some things in my use case (due to other issues), and didn't see any unit tests in the redis that I could expand on. If there's anything you think might be risky here (bug-wise) we can hold off.

I mean, if you can't validate this stuff, why are we considering this for inclusion?

bbaldino:
Sorry for the lack of follow-up here...we ended up going in a different direction (no longer using redis due to other reasons), so I likely won't have time to come back to this.

PR #161, opened 2023-05-23
Support redis::aio::MultiplexedConnection

jaymell:
The currently supported `redis::aio::Connection` has longstanding serious 
data inconsistency issues (e.g., https://github.com/redis-rs/redis-rs/issues/325). 
This change offers a ConnectionManager impl based on `MultiplexedConnection` 
as a safer alternative.

### DIFF
--- a/redis/src/lib.rs
+++ b/redis/src/lib.rs
@@ -39,7 +39,10 @@ pub use bb8;
 pub use redis;
 
 use async_trait::async_trait;
-use redis::{aio::Connection, ErrorKind};
+use redis::{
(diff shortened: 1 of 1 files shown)


jaymell:
We do want to get rid of the non-multiplexed connection, but regardless I don't think that will change the need for _something_ like this change here, i.e., I don't foresee us renaming `MultiplexedConnection`.

djc (approved):

djc:
Thanks!

djc:
I just published bb8-redis 0.13.1.

PR #162, opened 2023-05-24
Add LIFO option for getting pool resources

zainkabani:
The current strategy that bb8 uses for pulling from the connection queue is FIFO. This means that we are constantly resetting idle timers for connections. This PR adds an option to make the queue a stack so that the pool can reuse the most recently used resources and allow older ones to idle timeout.

### DIFF
--- a/bb8/src/api.rs
+++ b/bb8/src/api.rs
@@ -74,6 +74,14 @@ impl<M: ManageConnection> Pool<M> {
     }
 }
 
+#[derive(Debug)]
+pub enum QueueStrategy {
+    /// Last in first out
+    LIFO,
+    /// First in first out
+    FIFO,
+}
+
 /// A builder for a connection pool.
 #[derive(Debug)]
 pub struct Builder<M: ManageConnection> {
@@ -95,6 +103,8 @@ pub struct Builder<M: ManageConnection> {
     pub(crate) error_sink: Box<dyn ErrorSink<M::Error>>,
     /// The time interval used to wake up and reap connections.
     pub(crate) reaper_rate: Duration,
+    /// Queue strategy (FIFO or LIFO)
+    pub(crate) queue_strategy: QueueStrategy,
     /// User-supplied trait object responsible for initializing connections
     pub(crate) connection_customizer: Option<Box<dyn CustomizeConnection<M::Connection, M::Error>>>,
     _p: PhantomData<M>,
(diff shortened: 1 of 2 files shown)


bb8/src/api.rs
```diff
@@ -260,6 +271,15 @@ impl<M: ManageConnection> Builder<M> {
         self
     }
 
+    /// Sets the queue strategy to be used by the pool
+    ///
+    /// Defaults to `LIFO`.
```
dat2, line 276:
```suggestion
    /// Defaults to `FIFO`.
```
according to the code above

bb8/src/api.rs
```diff
@@ -74,6 +74,14 @@ impl<M: ManageConnection> Pool<M> {
     }
 }
 
+#[derive(Debug)]
+pub enum QueueStrategy {
+    /// Last in first out
+    LIFO,
```
djc, line 80: We should spell the variant name `Lifo` instead of `LIFO` per the naming guidelines. IMO the comment in its current state isn't that useful, I think it would be helpful to elaborate on the intended effect.

Also, please move the type definition below the `Builder` impl -- the prevailing style is sort of top-down.

I think it would be useful to derive `Default` on `QueueStrategy`, and then initialize `Builder::queue_strategy` to `QueueStrategy::default()`.

bb8/src/internals.rs
```diff
@@ -80,8 +81,14 @@ where
         }
 
         // Queue it in the idle queue
-        self.conns
-            .push_back(IdleConn::from(guard.conn.take().unwrap()));
+        match self.queue_strategy {
```
djc, line 84: I'd prefer to deduplicate the `IdleConn::from(guard.conn.take().unwrap())`.

djc (commented):
I'm open to merging this, but I have a bunch of stylistic suggestions.

bb8/src/api.rs
zainkabani: Thanks for the suggestions, I've implemented them in my latest changes but still failing the build tests. Can't reproduce on my machine so let me know if you have any ideas on how to pass them

djc: That's right -- I said to derive `Default` but that's not supported yet by the MSRV. Can you just add a custom `Default` `impl` that does the same thing instead? I like to avoid bumping the MSRV where possible.

zainkabani:
@djc Does this seem ready for merge on your end?

djc (approved):

djc:
Yup, thanks!

PR #163, opened 2023-05-26
Fix PoolInternals QueueStrategy default

zainkabani:
Fixing default for queue strategy


zainkabani:
@djc any plans to cut a release for bb8 soon? We're actively using it in our project and would prefer to be in sync with cargo rather than use github.

djc (approved):

PR #165, opened 2023-05-30
Fix configuration of queue strategy

zainkabani:
Pervious implementation of this feature was incorrect, this PR fixes it.


djc (approved):

djc:
I just published bb8 0.8.1.

PR #183, opened 2023-12-22
Add `redis-rs` feature flags

### LINKED ISSUE #146: Why bb8-redis can't support redis features?
I want use redis script but for bb8-redis features is disable `default-features = false`
How i can enabled them?

edmondwinston:
Fixes #146.

### DIFF
--- a/redis/Cargo.toml
+++ b/redis/Cargo.toml
@@ -10,8 +10,36 @@ repository = "https://github.com/djc/bb8"
 [dependencies]
 async-trait = "0.1"
 bb8 = { version = "0.8", path = "../bb8" }
-redis = { version = "0.24", default-features = false, features = ["tokio-comp"] }
+redis = { version = "0.24" }
+# redis = { version = "0.24", default-features = false, features = [
+#   "tokio-comp",
+# ] }
 
 [dev-dependencies]
 futures-util = "0.3.15"
 tokio = { version = "1.0.0", features = ["macros", "rt-multi-thread"] }
+
+[features]
+default = ["redis/default"]
+acl = []
+aio = ["redis/aio"]
+geospatial = []
+json = ["redis/json"]
+cluster = ["redis/cluster"]
+script = ["redis/script"]
+tls-native-tls = ["redis/tls-native-tls"]
+tls-rustls = ["redis/tls-rustls"]
+tls-rustls-insecure = ["redis/tls-rustls-insecure"]
+tls-rustls-webpki-roots = ["redis/tls-rustls-webpki-roots"]
+async-std-comp = ["redis/async-std-comp"]
+async-std-native-tls-comp = ["redis/async-std-native-tls-comp"]
+async-std-rustls-comp = ["redis/async-std-rustls-comp"]
+tokio-comp = ["redis/tokio-comp"]
+tokio-native-tls-comp = ["redis/tokio-native-tls-comp"]
+tokio-rustls-comp = ["redis/tokio-rustls-comp"]
+connection-manager = ["redis/connection-manager"]
+streams = []
+cluster-async = ["redis/cluster-async"]
+keep-alive = ["redis/keep-alive"]
+sentinel = ["redis/sentinel"]
+tcp_nodelay = []


edmondwinston:
@djc 

I've understood and used the first option, which is to have both `bb8-redis` and `redis-rs`, but (I think) this is still a nice think to have. However, I'm not sure this is what you meant, since I couldn't find a way to test it locally. Any suggestions?

<details>
  <summary>Metadata error</summary>
  
Error: failed to get cargo metadata

Caused by:
    `cargo metadata` exited with an error: warning: patch for `bb8-redis` uses the features mechanism. default-features and features will not take effect because the patch dependency does not support this mechanism
    error: failed to load source for dependency `bb8-redis`

    Caused by:
      Unable to update /home/edw/work/local-patches/bb8/redis

    Caused by:
      failed to read `/home/edw/work/local-patches/bb8/redis/Cargo.toml`

    Caused by:
      No such file or directory (os error 2)
</details>

redis/Cargo.toml
```diff
@@ -10,8 +10,36 @@ repository = "https://github.com/djc/bb8"
...
+#   "tokio-comp",
+# ] }
 
 [dev-dependencies]
 futures-util = "0.3.15"
 tokio = { version = "1.0.0", features = ["macros", "rt-multi-thread"] }
+
+[features]
+default = ["redis/default"]
+acl = []
```
djc, line 24: Any features added here that don't enable anything in the upstream redis crate just don't do anything, so we should not add them (or, I guess you meant `acl = ["redis/acl"]`?).

redis/Cargo.toml
```diff
@@ -10,8 +10,36 @@ repository = "https://github.com/djc/bb8"
 [dependencies]
 async-trait = "0.1"
 bb8 = { version = "0.8", path = "../bb8" }
-redis = { version = "0.24", default-features = false, features = ["tokio-comp"] }
+redis = { version = "0.24" }
```
djc, line 13: We'll still want the `default-features = false` here, I think, as the upstream crate's `default` features will then be re-enabled by our forwarding of our `default` feature to `redis/default`. I forget how or why redis has the `tokio-comp` feature but I also feel like we should keep that enabled unconditionally.

redis/Cargo.toml
```diff
@@ -10,8 +10,36 @@ repository = "https://github.com/djc/bb8"
...
+json = ["redis/json"]
+cluster = ["redis/cluster"]
+script = ["redis/script"]
+tls-native-tls = ["redis/tls-native-tls"]
+tls-rustls = ["redis/tls-rustls"]
+tls-rustls-insecure = ["redis/tls-rustls-insecure"]
+tls-rustls-webpki-roots = ["redis/tls-rustls-webpki-roots"]
+async-std-comp = ["redis/async-std-comp"]
+async-std-native-tls-comp = ["redis/async-std-native-tls-comp"]
+async-std-rustls-comp = ["redis/async-std-rustls-comp"]
```
djc, line 36: Given that bb8 unconditionally relies on tokio, I think we should not expose the `async-std` features here.

redis/Cargo.toml
```diff
@@ -10,8 +10,36 @@ repository = "https://github.com/djc/bb8"
...
+tokio-comp = ["redis/tokio-comp"]
+tokio-native-tls-comp = ["redis/tokio-native-tls-comp"]
+tokio-rustls-comp = ["redis/tokio-rustls-comp"]
```
djc, line 39: Given that we unconditionally enable `tokio-comp`, I think we should not expose these features here, and we can instead include their constituent parts in the `tls-*` features that we expose (so for example, `tls-rustls = ["redis/tls-rustls", "redis/tokio-rustls"]`).

djc:
> `cargo metadata` exited with an error: warning: patch for `bb8-redis` uses the features mechanism. default-features and features will not take effect because the patch dependency does not support this mechanism

That is just an aspect of how you've set this up. You can use features with a `patch` section, but features aren't configured as part of the `patch` section -- they're just taken from the `dependencies` section as usual.

redis/Cargo.toml
edmondwinston: Change made

redis/Cargo.toml
edmondwinston: Change made

redis/Cargo.toml
edmondwinston: Change made

redis/Cargo.toml
edmondwinston: Yes, I meant the latter - I wasn't cautious. Change made!

edmondwinston:
> You can use features with a `patch` section, but features aren't configured as part of the `patch` section -- they're just taken from the `dependencies` section as usual.

I see. Was doing it that way because I didn't know this is the way to expose upstream's features, so I very naturally want to test it locally, but I guess it is not necessary, after all.

redis/Cargo.toml
```diff
@@ -10,8 +10,29 @@ repository = "https://github.com/djc/bb8"
 [dependencies]
 async-trait = "0.1"
 bb8 = { version = "0.8", path = "../bb8" }
-redis = { version = "0.24", default-features = false, features = ["tokio-comp"] }
+redis = { version = "0.24", default-features = false, features = [
```
djc, line 13: Please revert the formatting change.

redis/Cargo.toml
```diff
@@ -10,8 +10,29 @@ repository = "https://github.com/djc/bb8"
...
+  "tokio-comp",
+] }
 
 [dev-dependencies]
 futures-util = "0.3.15"
 tokio = { version = "1.0.0", features = ["macros", "rt-multi-thread"] }
+
+[features]
```
djc, line 21: Please move the `features` before the `dependencies` section.

djc (approved):
This is almost ready!

PR #185, opened 2024-01-19
Switched to Into<Option<T>> for parameters in Builder

fpagliughi:
Hey, I know this is minor, but it seems so much nicer for optional parameters in a public API to take `Into<Option<T>>` rather than just taking `Option<T>`. That way the user can just send a value of `T` for `Some<T>`, and of course can still send `None` when desired. 

Thus we could do:
```
Builder::new().min_idle(2)...
```
rather than
```
Builder::new().min_idle(Some(2))...
```
but, `min_idle(Some(2))` and `min_idle(None)` still work, so it's all backward compatible.

### DIFF
--- a/bb8/src/api.rs
+++ b/bb8/src/api.rs
@@ -165,10 +165,7 @@ impl<M: ManageConnection> Builder<M> {
     ///
     /// Defaults to None.
     #[must_use]
-    pub fn min_idle<U>(mut self, min_idle: U) -> Self
-    where
-        U: Into<Option<u32>>,
-    {
+    pub fn min_idle(mut self, min_idle: impl Into<Option<u32>>) -> Self {
         self.min_idle = min_idle.into();
         self
     }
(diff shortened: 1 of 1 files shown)


bb8/src/api.rs
```diff
@@ -165,8 +165,11 @@ impl<M: ManageConnection> Builder<M> {
     ///
     /// Defaults to None.
     #[must_use]
-    pub fn min_idle(mut self, min_idle: Option<u32>) -> Self {
-        self.min_idle = min_idle;
+    pub fn min_idle<U>(mut self, min_idle: U) -> Self
```
djc, line 168: Nit: if we're going to do this, I'd like it spelled `min_idle: impl Into<Option<u32>>` (same for the others).

djc (approved):
Seems nice, thanks!

bb8/src/api.rs
fpagliughi: Oh, I had never done it like that, but actually that looks pretty good.

djc:
Thanks!

PR #189, opened 2024-02-01
Track PooledConnection state explicitly

djc:


djc:
Merged and published this proactively to make this easy to consume and since I'm pretty sure it addresses a regression from 0.8.2. Still happy to address any feedback anyone might have!

PR #191, opened 2024-02-02
Remove dependency on `futures-channel` in favour of `tokio` primitives

aumetra:
Since `bb8` is already depending on Tokio, it doesn't really make sense for it to pull in `futures-channel`.  
Especially since this dependency was only used for tests.


djc (approved):
Nice catch!

PR #194, opened 2024-03-25
Fix #167: Notify waiters when dropping a bad connection from the pool

### LINKED ISSUE #167: TimedOut error when has_broken = true
Hi,

I found what seems to be a bug. When multiple connections return `true` on `has_broken()`, all future requests to `get()` fail with a `TimedOut` error. This code can be used to reproduce the problem:

```
use std::{convert::Infallible, time::Duration};

struct ConnectionManager;
struct Connection;

#[async_trait::async_trait]
impl bb8::ManageConnection for ConnectionManager {
    type Connection = Connection;
    type Error = Infallible;

        Ok(Connection)
    }

    async fn is_valid(&self, _: &mut Self::Connection) -> Result<(), Self::Error> {
        Ok(())
    }

    fn has_broken(&self, _: &mut Self::Connection) -> bool {
        true
    }
}

#[tokio::test]
async fn pool() {
    let pool = bb8::Pool::builder()
        .max_size(5)
        .connection_timeout(Duration::from_secs(10))
        .build(ConnectionManager)
        .await
        .unwrap();

    let mut futures = Vec::new();

    for _ in 0..10 {
        let pool = pool.clone();
        futures.push(tokio::spawn(async move {
            drop(conn);
        }));
    }

    for future in futures {
        future.await.unwrap();
    }
}

```

xortive:


djc (approved):
This looks reasonable, thanks!

Can you clean up the warnings from your test code as pointed out by CI?

djc (approved):

djc:
I'm still digging into your code and wondering if `put_back()` is the best place to notify the waiters? You mentioned that we should also do it in `Reaper::run()` -- so maybe we should move it into `spawn_replenishing_approvals()`? I guess that would wake up waiters in the `PoolInner::get()` loop, too, which might be weird.

In the end I wonder why `notify_one()` in `PoolInternals::put()` isn't enough -- that's called by both `put_back()` (for the success case) and `add_connection()`, which is ultimately what gets called during the replenishing process.

djc:
> The problem that this fixed for the particular test case provided, was all about the failure case in `put_back()`. When a bad connection is returned to the pool, while someone is waiting for a good connection, they still need to be notified of an available spot in the pool so they can go ahead and create a new connection to replace it.

That doesn't quite make sense to me because in the `put_back()` failure case we already:

- Called `spawn_replenishing_approvals()`, which
- Calls `replenish_idle_connections()`, which
- Calls `add_connections()`, which
- Calls `add_connection()`, which
- Calls `PoolInternals::put()`, which
- Calls `pool.notify.notify_one()`

So why isn't that sufficient?

vaikzs:
@djc I noticed that the final call to `spawn_replenishing_approvals(approvals)` is not complete due to empty approvals and returns because there are 5 connections waiting. Due to this, call stack doesn't reach `pool.notify.notify_one()`. 

However, instead of adding `pool.notify.notify_waiters()` in `put_back()` which makes it redundant, replacing `pool.notify.notify_one()` in `PoolInternals::put` to `pool.notify.notify_waiters()` (no permit is stored to be used by the next call to `notified().await`) solves this. I tested with the same test case provided by @xortive
```rust
#[tokio::test]
async fn test_broken_connections_dont_starve_pool() {
    use std::sync::RwLock;
    use std::{convert::Infallible, time::Duration};

    #[derive(Default)]
    struct ConnectionManager {
        counter: RwLock<u16>,
    }
    #[derive(Debug)]
    struct Connection;

    #[async_trait::async_trait]
    impl bb8::ManageConnection for ConnectionManager {
        type Connection = Connection;
        type Error = Infallible;

            Ok(Connection)
        }

        async fn is_valid(&self, _: &mut Self::Connection) -> Result<(), Self::Error> {
            Ok(())
        }

        fn has_broken(&self, _: &mut Self::Connection) -> bool {
            let mut counter = self.counter.write().unwrap();
            let res = *counter < 5;
            *counter += 1;
            res
        }
    }

    let pool = bb8::Pool::builder()
        .max_size(5)
        .connection_timeout(Duration::from_secs(10))
        .build(ConnectionManager::default())
        .await
        .unwrap();

    let mut futures = Vec::new();

    for _ in 0..10 {
        let pool = pool.clone();
        futures.push(tokio::spawn(async move {
            drop(conn);
        }));
    }

    for future in futures {
        future.await.unwrap();
    }
}
```

djc:
@vaikzs thanks for digging in! Would you be able to submit a PR, ideally including a test that tickles this behavior?

PR #198, opened 2024-04-18
Provide statistics for get operations

pfreixes:
# Exposing three new attributes as part of a new `Statistics` type
We will be exposing the three new attributes as part of a new `Statistics` type:

- **get_direct**: Total number of `get` operations performed for retrieving a new connection that did no wait for getting a connection.
- **get_waited**: Total number of `get` operations that had to wait for having a connection available.
- **get_timed_out**: Total number of `get` operations that did time out while waiting for a connection.

See following section for understanding how these two values can be used

# Why adding these three new attributes ?

Tuning the connection pool, specifically the max size and the min idle, its something that needs to be done by gathering data that helps you to understand how well configure are for example these to configuration parameters, which should lead to an ideal situation where the contention of the pool is affordable.

By having visibility of the gets that managed to retrieve a connection without waiting and having too the visibility of the ones that had to wait, the user would have available enough data for understanding if the pool is having contention.

## When contention can happen

There are multiple scenarios where contention at connection pool level can happen, which can be described as the scenario where there are no enough connections free in the pool and the calls for getting a connection need to wait

- When the max size of the pool was underestimated, and the operations for getting a new connection need to constantly wait for having a connection free.
- When the max idle connection is too low, and the operations for getting a new connection have multiple chances to have to pay the cost of creating a new connection.

By adding the `get_direct`, `get_waited` and `get_timed_out` we provide a way to know deterministically  how much contention the connection pool is suffering. By using a simple operation like `(get_waited+get_timed_out)/(get_direct + get_waited+get_timed_out)` the user can understand the percentage of operations that suffered from connection pooling contention.

A low number - i.e < 1% - might indicate that the two variables are properly configured, while bigger numbers - i.e > 5% - might indicate that there is a none negligible amount of operations that are suffering an extra latency due to some level of contention when getting a connection.

# How and when to use it?

There is an expectation that the `State` will be polled every X times, and the user will take the responsibility of calculating the delta from the previous values, something like:

```rust
let state_before = pool.statistics()
let state_after = pool.statistics()

println!("Gets performed within period X: {}", (state_after.get_direct + state_after.get_waited + state_after.get_timed_out) - (state_before.get_direct + state_after.get_waited + state_after.get_timed_out))
println!("Gets performed within period X that wait for a connection: {}", state_after.get_waited - state_before.get_waited)
println!("Gets performed within period X that wait for a connection and timed out: {}", state_after.get_timed_out - state_before.get_timed_out)
```

Deltas can be then sent to your favourite time series database/service - i.e prometheus, DD, etc.

### DIFF
--- a/bb8/src/api.rs
+++ b/bb8/src/api.rs
@@ -85,6 +85,20 @@ pub struct State {
     pub connections: u32,
     /// The number of idle connections.
     pub idle_connections: u32,
+    /// Statistics about the historical usage of the pool.
+    pub statistics: Statistics,
+}
+
+/// Statistics about the historical usage of the `Pool`.
+#[derive(Debug, Default)]
+#[non_exhaustive]
+pub struct Statistics {
+    /// Total gets performed that did not have to wait for a connection.
+    pub get_direct: u64,
+    /// Total gets performed that had to wait for a connection available.
+    pub get_waited: u64,
+    /// Total gets performed that timed out while waiting for a connection.
+    pub get_timed_out: u64,
 }
 
 /// A builder for a connection pool.
--- a/bb8/src/inner.rs
+++ b/bb8/src/inner.rs
@@ -10,7 +10,7 @@ use tokio::spawn;
 use tokio::time::{interval_at, sleep, timeout, Interval};
 
 use crate::api::{Builder, ConnectionState, ManageConnection, PooledConnection, RunError, State};
-use crate::internals::{Approval, ApprovalIter, Conn, SharedPool};
+use crate::internals::{Approval, ApprovalIter, Conn, SharedPool, StatsKind};
 
 pub(crate) struct PoolInner<M>
 where
@@ -85,6 +85,8 @@ where
     }
 
     pub(crate) async fn get(&self) -> Result<PooledConnection<'_, M>, RunError<M::Error>> {
+        let mut kind = StatsKind::Direct;
+
         let future = async {
             loop {
                 let (conn, approvals) = self.inner.pop();
@@ -96,6 +98,7 @@ where
                 let mut conn = match conn {
                     Some(conn) => PooledConnection::new(self, conn),
                     None => {
+                        kind = StatsKind::Waited;
                         self.inner.notify.notified().await;
                         continue;
                     }
@@ -116,10 +119,16 @@ where
             }
         };
 
-        match timeout(self.inner.statics.connection_timeout, future).await {
+        let result = match timeout(self.inner.statics.connection_timeout, future).await {
             Ok(result) => result,
-            _ => Err(RunError::TimedOut),
-        }
+            _ => {
+                kind = StatsKind::TimedOut;
+                Err(RunError::TimedOut)
+            }
+        };
+
+        self.inner.statistics.record(kind);
+        result
     }
 
     pub(crate) async fn connect(&self) -> Result<M::Connection, M::Error> {
@@ -148,7 +157,10 @@ where
 
     /// Returns information about the current state of the pool.
     pub(crate) fn state(&self) -> State {
-        (&*self.inner.internals.lock()).into()
+        self.inner
+            .internals
+            .lock()
+            .state((&self.inner.statistics).into())
     }
 
     // Outside of Pool to avoid borrow splitting issues on self
--- a/bb8/src/internals.rs
+++ b/bb8/src/internals.rs
@@ -1,12 +1,13 @@
 use std::cmp::min;
+use std::collections::VecDeque;
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
 use std::time::Instant;
 
-use crate::{api::QueueStrategy, lock::Mutex};
 use tokio::sync::Notify;
 
-use crate::api::{Builder, ManageConnection, State};
-use std::collections::VecDeque;
+use crate::api::{Builder, ManageConnection, QueueStrategy, State, Statistics};
+use crate::lock::Mutex;
 
 /// The guts of a `Pool`.
 #[allow(missing_debug_implementations)]
@@ -18,6 +19,7 @@ where
     pub(crate) manager: M,
     pub(crate) internals: Mutex<PoolInternals<M>>,
     pub(crate) notify: Arc<Notify>,
+    pub(crate) statistics: AtomicStatistics,
 }
 
 impl<M> SharedPool<M>
@@ -30,6 +32,7 @@ where
             manager,
             internals: Mutex::new(PoolInternals::default()),
             notify: Arc::new(Notify::new()),
+            statistics: AtomicStatistics::default(),
         }
     }
 
@@ -153,14 +156,12 @@ where
 
         self.dropped((before - self.conns.len()) as u32, config)
     }
-}
 
-#[allow(clippy::from_over_into)] // Keep this more private with the internal type
-impl<M: ManageConnection> Into<State> for &PoolInternals<M> {
-    fn into(self) -> State {
+    pub(crate) fn state(&self, statistics: Statistics) -> State {
         State {
             connections: self.num_conns,
             idle_connections: self.conns.len() as u32,
+            statistics,
         }
     }
 }
@@ -247,4 +248,37 @@ impl<C: Send> From<Conn<C>> for IdleConn<C> {
             idle_start: Instant::now(),
         }
     }
+}
+
+#[derive(Default)]
+pub(crate) struct AtomicStatistics {
+    pub(crate) get_direct: AtomicU64,
+    pub(crate) get_waited: AtomicU64,
+    pub(crate) get_timed_out: AtomicU64,
+}
+
+impl AtomicStatistics {
+    pub(crate) fn record(&self, kind: StatsKind) {
+        match kind {
+            StatsKind::Direct => self.get_direct.fetch_add(1, Ordering::SeqCst),
+            StatsKind::Waited => self.get_waited.fetch_add(1, Ordering::SeqCst),
+            StatsKind::TimedOut => self.get_timed_out.fetch_add(1, Ordering::SeqCst),
+        };
+    }
+}
+
+impl From<&AtomicStatistics> for Statistics {
+    fn from(item: &AtomicStatistics) -> Self {
+        Self {
+            get_direct: item.get_direct.load(Ordering::SeqCst),
+            get_waited: item.get_waited.load(Ordering::SeqCst),
+            get_timed_out: item.get_timed_out.load(Ordering::SeqCst),
+        }
+    }
+}
+
+pub(crate) enum StatsKind {
+    Direct,
+    Waited,
+    TimedOut,
 }
--- a/bb8/tests/test.rs
+++ b/bb8/tests/test.rs
@@ -885,3 +885,43 @@ async fn test_broken_connections_dont_starve_pool() {
         future.await.unwrap();
     }
 }
+
+#[tokio::test]
+async fn test_state_get_contention() {
+    let pool = Pool::builder()
+        .max_size(1)
+        .min_idle(1)
+        .build(OkManager::<FakeConnection>::new())
+        .await
+        .unwrap();
+
+    let (tx1, rx1) = oneshot::channel();
+    let (tx2, rx2) = oneshot::channel();
+    let clone = pool.clone();
+    tokio::spawn(async move {
+        let conn = clone.get().await.unwrap();
+        tx1.send(()).unwrap();
+        let _ = rx2
+            .then(|r| match r {
+                Ok(v) => ok((v, conn)),
+                Err(_) => err((Error, conn)),
+            })
+            .await;
+    });
+
+    // Get the first connection.
+    rx1.await.unwrap();
+
+    // Now try to get a new connection without waiting.
+    let f = pool.get();
+
+    // Release the first connection.
+    tx2.send(()).unwrap();
+
+    // Wait for the second attempt to get a connection.
+    f.await.unwrap();
+
+    let statistics = pool.state().statistics;
+    assert_eq!(statistics.get_direct, 1);
+    assert_eq!(statistics.get_waited, 1);
+}


djc:
I'm okay with adding some statistics to bb8, but this PR is pretty far from something I'd like to merge. Please concisely describe your use cases -- which information you need and why -- and then be prepared to invest time to very carefully follow any guidance I provide on how to implement this.

pfreixes:
> Please concisely describe your use cases -- which information you need and why --

Tried to update the PR with a much better explanation, please leat me know if you believe that needs still some work

> and then be prepared to invest time to very carefully follow any guidance I provide on how to implement this.

That would be awesome 🙇

djc:
> > Please concisely describe your use cases -- which information you need and why --
> 
> Tried to update the PR with a much better explanation, please leat me know if you believe that needs still some work

Can you give some (semi-) formal definitions for conceptually what values you want to expose?

bb8/src/inner.rs
```diff
@@ -10,13 +11,36 @@ use tokio::spawn;
 use tokio::time::{interval_at, sleep, timeout, Interval};
 
 use crate::api::{Builder, ConnectionState, ManageConnection, PooledConnection, RunError};
-use crate::internals::{Approval, ApprovalIter, Conn, SharedPool, State};
+use crate::internals::{Approval, ApprovalIter, Conn, SharedPool};
+
+struct SharedPoolInnerStats {
```
djc, line 16: Why can't this be a part of the `SharedPool`?

Style: code in bb8 follows a top-down style, so definitions should be moved below any code that is using it (in descending order of size/complexity).

bb8/src/internals.rs
```diff
@@ -249,7 +249,7 @@ impl<C: Send> From<Conn<C>> for IdleConn<C> {
 /// Information about the state of a `Pool`.
 #[derive(Debug)]
 #[non_exhaustive]
-pub struct State {
+pub struct PoolInternalState {
```
djc, line 252: ... and leave this name as is.

bb8/src/inner.rs
pfreixes: > Why can't this be a part of the SharedPool?

I guess that was the place were was easily to measure the [`get`](https://github.com/pfreixes/bb8/blob/add_get_stats/bb8/src/inner.rs#L116) contention. Also if we move forward with this PR I would like to also add new attribute for counting the total time waited, something similar to the [`WaitDuration`](https://pkg.go.dev/database/sql#DBStats) exposed by Golang for the `sql.DB` pool. If we go in that way for making this calculation the `SharedPool` will give us the right place for measuring this time, 

> Style: code in bb8 follows a top-down style, so definitions should be moved below any code that is using it (in descending order of size/complexity).
Oks Ill move this to down

bb8/src/inner.rs
```diff
@@ -244,3 +294,22 @@ impl<M: ManageConnection> Reaper<M> {
         }
     }
 }
+
+/// Information about the state of a `Pool`.
+#[derive(Debug)]
+#[non_exhaustive]
+pub struct State {
```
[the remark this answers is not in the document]
pfreixes: Oks, do we want to keep returning this as part of the `get_state` or should we crate a new `get_statistics` method? As a similar name we could just say it `stats`

bb8/src/internals.rs
pfreixes: 👍

bb8/src/inner.rs
pfreixes: Also we might want to consider that other attributes like `MaxIdleClosed`, `MaxIdleTimeClosed`, `MaxLifetimeClosed`.

bb8/src/inner.rs
```diff
@@ -10,13 +11,36 @@ use tokio::spawn;
 use tokio::time::{interval_at, sleep, timeout, Interval};
 
 use crate::api::{Builder, ConnectionState, ManageConnection, PooledConnection, RunError};
-use crate::internals::{Approval, ApprovalIter, Conn, SharedPool, State};
+use crate::internals::{Approval, ApprovalIter, Conn, SharedPool};
+
+struct SharedPoolInnerStats {
+    gets: AtomicU32,
+    gets_waited: AtomicU32,
```
pfreixes, line 18: @djc I was thinking that just for the sake to reduce the number of events for having overflows and resets to 0, we could just use `AtomicU64`.

djc: Sounds good.

bb8/src/inner.rs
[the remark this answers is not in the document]
djc: Probably a `statistics()` method.

bb8/src/inner.rs
djc: > I guess that was the place were was easily to measure the [get](https://github.com/pfreixes/bb8/blob/add_get_stats/bb8/src/inner.rs#L116) contention.

What do you mean? What does the place to measure have to do with the place for storing?

pfreixes: > What do you mean? What does the place to measure have to do with the place for storing?

🤦 now i see that I totally digressed to a different subject, sorry. I just stored it into a different class without any specific reason, Im happy to move them as simple attributes of the `SharedPool` and thats all.

pfreixes: Now Ive remembered another reason for having all new statistics attributes under its own type, if I do not do this this would force me to warp all of them under the `Arc` object.

If this is not a big issue for you I can do this, otherwise with current proposal we can just share a reference to one single instance of `SharedPoolInnerStatistics`.

Up to you

pfreixes:
@djc PR should be ready for another review?

I would like to work on extending the statistics for adding too the following ones:

- WaitDuration    Total time waited for a connection
- MaxIdleClosed  Connections closed due to max idle
- MaxIdleTimeClosed Connections closed due to max idle time
- MaxLifetimeClosed Connections closed due to max time
- ClosedConnectionsWithError Connections closed due to errors
- ClosedConnections Total Connections closed
- OpenedConnections Total Connections opened

I can do this on top of this PR or in a new one, any objections? any preference? If we keep the `statistics` as public interface for gathering all fields, Im happy to go in either way, since we would be able to make internal breaking changes if required.

pfreixes:
Regarding my latest comment, just finished the [implementation](https://github.com/pfreixes/bb8/compare/add_get_stats...pfreixes:bb8:more_statistics?expand=1) of other statistics so up to you how you want to handle this second batch of changes.

Just let me know.

pfreixes:
Any new feedback on this one? 🙇

pfreixes:
Hi @djc hope that everything is fine, if you have time I would love to have some feedback with regards the outstanding questions for this PR 🙇

djc:
Yes, will try to follow up soon (was on vacation for two weeks).

bb8/src/inner.rs
```diff
@@ -116,10 +126,14 @@ where
...
         };
 
-        match timeout(self.inner.statics.connection_timeout, future).await {
+        let result = match timeout(self.inner.statics.connection_timeout, future).await {
             Ok(result) => result,
             _ => Err(RunError::TimedOut),
-        }
+        };
+
+        self.pool_inner_stats.record_get(with_contention);
```
djc, line 134: If this times out, should we still record a get, or is this arguably a third kind of event?

bb8/src/inner.rs
```diff
@@ -25,6 +27,7 @@ where
 {
     pub(crate) fn new(builder: Builder<M>, manager: M) -> Self {
         let inner = Arc::new(SharedPool::new(builder, manager));
+        let pool_inner_stats = Arc::new(SharedPoolInnerStatistics::new());
```
djc, line 30: Why can't `SharedPoolInnerStatistics` live inside the `SharedPool`?

bb8/src/inner.rs
```diff
@@ -244,3 +271,39 @@ impl<M: ManageConnection> Reaper<M> {
         }
     }
 }
+
+struct SharedPoolInnerStatistics {
```
djc, line 275: Let's name this `AtomicStatistics`.

bb8/src/inner.rs
```diff
@@ -146,6 +160,14 @@ where
         }
     }
 
+    /// Returns statistics about the historical usage of the pool.
+    pub(crate) fn statistics(&self) -> Statistics {
+        let gets = self.pool_inner_stats.gets.load(Ordering::SeqCst);
```
djc, line 165: Would prefer to isolate this into an `impl From<AtomicStatistics> for Statistics`.

bb8/src/inner.rs
```diff
@@ -244,3 +271,39 @@ impl<M: ManageConnection> Reaper<M> {
...
     }
 }
+
+struct SharedPoolInnerStatistics {
+    gets: AtomicU64,
+    gets_waited: AtomicU64,
+}
+
+impl SharedPoolInnerStatistics {
+    fn new() -> Self {
```
djc, line 281: Let's get rid of `new()` in favor of deriving `Default`.

bb8/src/inner.rs
```diff
@@ -244,3 +271,39 @@ impl<M: ManageConnection> Reaper<M> {
...
+        Self {
+            gets: AtomicU64::new(0),
+            gets_waited: AtomicU64::new(0),
+        }
+    }
+
+    fn record_get(&self, with_contention: bool) {
+        self.gets.fetch_add(1, Ordering::SeqCst);
+
+        if with_contention {
```
djc, line 291: It would be more efficient to do one `fetch_add()` per `record_get()`, such that we record into uncontended or contended gets (the other info can be derived).

bb8/src/inner.rs
pfreixes: IMO we should still record it, timeout is the just a safeguard for not having indefinitely polling for getting a connection, which means that suffered from the issues that this new statistic is trying to inform to.

True that the user that uses this statistic should not blindly believe that an increase of gets with contention is just an issue with the connection allocated in the pool, as you were mentioning timeout could also indicate that the server is for example having issues creating new connections (for whatever reasons).

I would not go for a time out event for now, unless someones believes that the addition of this one would address some lack on observability.

bb8/src/inner.rs
pfreixes: Just for curiosity - kinda new in Rust - is this just a pattern detail or has some implications?

bb8/src/inner.rs
pfreixes: So you mean, just making one single `fetch_add` for one or the other statistic, and when we return info about both of them we return the sum of both for returning the total gets?

bb8/src/inner.rs
djc: It's just less code because all the values here are the `Default` values for their respective types anyway. Less code is almost always better!

bb8/src/inner.rs
djc: I think the info we return might as well be the info we gather, and then the downstream caller can still aggregate those numbers to calculate this number.

bb8/src/inner.rs
pfreixes: Good point, thanks for the clarification.

bb8/src/inner.rs
djc: Hmm, I think it would be strictly more useful to have three fields on `AtomicStatistics`:

* `get_direct`
* `get_waited`
* `get_timed_out`

pfreixes: Oks, lets go for it

pfreixes:
@djc I believe that Ive addressed all of your comments PTAL (thanks for the review, right now the changes seem to be more aligned with the current base code!)

bb8/src/inner.rs
```diff
@@ -85,6 +86,8 @@ where
     }
 
     pub(crate) async fn get(&self) -> Result<PooledConnection<'_, M>, RunError<M::Error>> {
+        let mut get_direct = true;
```
djc, line 89: Nit: let's just call this `direct`.

bb8/src/inner.rs
```diff
@@ -117,8 +121,27 @@ where
         };
 
         match timeout(self.inner.statics.connection_timeout, future).await {
-            Ok(result) => result,
-            _ => Err(RunError::TimedOut),
+            Ok(result) => {
+                if get_direct {
+                    self.inner
```
djc, line 126: This is a little verbose for my taste. Suggest defining an `enum GetKind { Direct, Indirect, TimedOut }` and a method `SharedPool::record(kind: GetKind)` that abstracts over the details here.

Also I think this is missing records for the direct path around line 109 and 113?

bb8/src/inner.rs
```diff
@@ -244,3 +272,26 @@ impl<M: ManageConnection> Reaper<M> {
         }
     }
 }
+
+/// Statistics about the historical usage of the `Pool`.
+#[derive(Debug)]
+#[non_exhaustive]
+pub struct Statistics {
+    /// Information about gets.
```
djc, line 280: Let's remove this line? Doesn't seem worth it.

bb8/src/api.rs
```diff
@@ -45,6 +46,11 @@ impl<M: ManageConnection> Pool<M> {
         Builder::new()
     }
 
+    /// Returns statistics about the historical usage of the pool.
+    pub fn statistics(&self) -> Statistics {
```
djc, line 50: Hmm, actually I think we should make `statistics` a field in `State`. What do you think?

djc (commented):
Much better, getting pretty close!

bb8/src/inner.rs
pfreixes: > This is a little verbose for my taste. Suggest defining an enum GetKind { Direct, Indirect, TimedOut } and a method SharedPool::record(kind: GetKind) that abstracts over the details here.

Yeps I can do that but this verbosity will be in some way unavoidable, within the `record` method we will need to branch for increasing one of the atomics, true that we can just use a `match` operator that could make this verbosity more shinny.

> Also I think this is missing records for the direct path around line 109 and 113?

It should not unless Im mistaken, in the end we are waiting for either the future to resolve or the timeout, so any result from the future if the timeout is not triggered will end up either as a `get_direct` or as a `get_waited`. Or am I missing something here?

bb8/src/api.rs
pfreixes: I would expect that for having a full representation of all metrics, the ones coming from the `State` would end us gauge metrics, the user will need to gather both objects, so Im ok to make this easier to the user by only requiring to make a single api call for gathering both.

Ill do that.

bb8/src/inner.rs
djc: > Yeps I can do that but this verbosity will be in some way unavoidable, within the `record` method we will need to branch for increasing one of the atomics, true that we can just use a `match` operator that could make this verbosity more shinny.

Sure -- but `get()` is already pretty complex so it would be nice to move some of the complexity elsewhere. And it should be possible to write `record()` in a way that it doesn't need to repeat the `SeqCst` 3 times.

> > Also I think this is missing records for the direct path around line 109 and 113?
> 
> It should not unless Im mistaken, in the end we are waiting for either the future to resolve or the timeout, so any result from the future if the timeout is not triggered will end up either as a `get_direct` or as a `get_waited`. Or am I missing something here?

No, you're right! Sorry, was thinking of an earlier version of this code.

pfreixes: Oki doki, Ill make the changes later!

pfreixes:
Hi @djc i did the changes related to moving the verbose logic for recording the gets in this commit https://github.com/djc/bb8/pull/198/commits/bc9fba48de34c1e796caa58495861cf47a0c4ce1 PTAL

I did not go for now on adding `Statistics` as part of the `State` object, I wanted to first gather your thoughts, if we do that movement we might be forced to move the `State` object as part of the inner module and create a new internal like `StatePoolnternal` that would replace the former `State` internals, once this is done we would be doing the composition of the object into the `PoolInner::state()` method.  The other option would be to move the `AtomicStatistics` under the `PoolInternals` but considering that this is an object guarded by a Mutex and with a very specific semantics, I would not go in that way.

The third option is just keep the current implementation with two segregate interfaces.

Thoughts?

djc:
> I did not go for now on adding `Statistics` as part of the `State` object, I wanted to first gather your thoughts, if we do that movement we might be forced to move the `State` object as part of the inner module and create a new internal like `StatePoolnternal` that would replace the former `State` internals, once this is done we would be doing the composition of the object into the `PoolInner::state()` method. The other option would be to move the `AtomicStatistics` under the `PoolInternals` but considering that this is an object guarded by a Mutex and with a very specific semantics, I would not go in that way.

I think the changes in #200 would enable doing it the way I suggested, do you agree?

pfreixes:
> I think the changes in https://github.com/djc/bb8/pull/200 would enable doing it the way I suggested, do you agree

Yeps Ill piggyback on your changes for making mine changes, thanks for taking care of it!

bb8/src/inner.rs
```diff
@@ -148,7 +158,9 @@ where
 
     /// Returns information about the current state of the pool.
     pub(crate) fn state(&self) -> State {
-        (&*self.inner.internals.lock()).into()
+        let mut state: State = (&*self.inner.internals.lock()).into();
+        state.statistics = (&(self.inner.statistics)).into();
+        state
```
pfreixes, line 163: I do very dislike very much what i had to here 😓 

So either you know how to do that this better, if not my porposal if we wan to keep this aggregated is to just use three different object, `State`, `PoolState` and `Statistics`. The `State` one would have the instances of the `PoolState` and `Statistics`.

Thoughts @djc ?

pfreixes: @djc pinging you kere for when yo have a moment 🙇

djc: Took the liberty of pushing to your branch. Let's see what CI thinks.

pfreixes (commented):
LGTM!

djc:
Thanks for your patience, happy to see more statistics in your next PR!

pfreixes:
Thanks for the review process, helped me a lot with my journey for learning Rust.

PR #201, opened 2024-06-03
fix: Notification to waiting clients during drop

vaikzs:
Following up with @djc's intuition (re: https://github.com/djc/bb8/pull/194) -- I noticed that the final call to `spawn_replenishing_approvals(approvals)` is incomplete due to empty approvals and returns early after several attempts to `put_back` keeping 5 connections waiting. Due to this, the call stack doesn't reach `pool.notify.notify_one()` at the end, and the `pool.notify.notify_one()` invocation at every connection polling (i.e.) also fails. 

However, instead of adding `pool.notify.notify_waiters()` in `put_back()` which makes it redundant, replacing `pool.notify.notify_one()` in `PoolInternals::put` to `pool.notify.notify_waiters()` (no permit is stored to be used by the next call to `notified().await`) solves this. I tested with the same test case (shown below and already included in the main branch) provided by @xortive 

```rust
#[tokio::test]
async fn test_broken_connections_dont_starve_pool() {
    use std::sync::RwLock;
    use std::{convert::Infallible, time::Duration};

    #[derive(Default)]
    struct ConnectionManager {
        counter: RwLock<u16>,
    }
    #[derive(Debug)]
    struct Connection;

    #[async_trait::async_trait]
    impl bb8::ManageConnection for ConnectionManager {
        type Connection = Connection;
        type Error = Infallible;

            Ok(Connection)
        }

        async fn is_valid(&self, _: &mut Self::Connection) -> Result<(), Self::Error> {
            Ok(())
        }

        fn has_broken(&self, _: &mut Self::Connection) -> bool {
            let mut counter = self.counter.write().unwrap();
            let res = *counter < 5;
            *counter += 1;
            res
        }
    }

    let pool = bb8::Pool::builder()
        .max_size(5)
        .connection_timeout(Duration::from_secs(10))
        .build(ConnectionManager::default())
        .await
        .unwrap();

    let mut futures = Vec::new();

    for _ in 0..10 {
        let pool = pool.clone();
        futures.push(tokio::spawn(async move {
            drop(conn);
        }));
    }

    for future in futures {
        future.await.unwrap();
    }
}
```

### DIFF
--- a/bb8/src/inner.rs
+++ b/bb8/src/inner.rs
@@ -141,7 +141,6 @@ where
             (_, _) => {
                 let approvals = locked.dropped(1, &self.inner.statics);
                 self.spawn_replenishing_approvals(approvals);
-                self.inner.notify.notify_waiters();
             }
         }
     }
--- a/bb8/src/internals.rs
+++ b/bb8/src/internals.rs
@@ -95,7 +95,7 @@ where
             QueueStrategy::Lifo => self.conns.push_front(conn),
         }
 
-        pool.notify.notify_one();
+        pool.notify.notify_waiters();
     }
 
     pub(crate) fn connect_failed(&mut self, _: Approval) {


djc:
Sorry, I still don't completely understand why this is necessary. Why doesn't my call stack in https://github.com/djc/bb8/pull/194#issuecomment-2041180473 hold? If something is returning an empty approval iter in a state where that is not sufficient, why is that? Pointing at an existing test case doesn't help me because that test already passes today.

PR #202, opened 2024-06-05
Adds new statistics for tracking connections created and closed

pfreixes:
We add several new statistics fields for tracking the following:

- `connections_created`: How many connections have been created
- `connections_broken_closed`: How many connections were closed due to a broken state
- `connections_invalid_closed`: How many connections were closed due to be considered invalid.
- `connections_max_lifetime_closed`: How many connections were closed due to reaching the max lifetime.
- `connections_max_idle_timeout_closed`: How many connections were closed due to reaching the idle timeout.

We have added also a new statistic called `get_waited_time_micro` which provides the accumulated time in micros that all calls to the `get` method had to wait for a new connection.

### DIFF
--- a/bb8/src/api.rs
+++ b/bb8/src/api.rs
@@ -100,7 +100,7 @@ pub struct Statistics {
     /// Total gets performed that timed out while waiting for a connection.
     pub get_timed_out: u64,
     /// Total time accumulated waiting for a connection.
-    pub get_waited_time_micro: u64,
+    pub get_waited_time: Duration,
     /// Total connections created.
     pub connections_created: u64,
     /// Total connections that were closed due to be in broken state.
--- a/bb8/src/api.rs
+++ b/bb8/src/api.rs
@@ -99,6 +99,20 @@ pub struct Statistics {
     pub get_waited: u64,
     /// Total gets performed that timed out while waiting for a connection.
     pub get_timed_out: u64,
+    /// Total time accumulated waiting for a connection.
+    pub get_waited_time_micro: u64,
+    /// Total connections created.
+    pub connections_created: u64,
+    /// Total connections that were closed due to be in broken state.
+    pub connections_broken_closed: u64,
+    /// Total connections that were closed due to be considered invalid.
+    pub connections_invalid_closed: u64,
+    /// Total connections that were closed because they reached the max
+    /// lifetime.
+    pub connections_max_lifetime_closed: u64,
+    /// Total connections that were closed because they reached the max
+    /// idle timeout.
+    pub connections_max_idle_timeout_closed: u64,
 }
 
 /// A builder for a connection pool.
--- a/bb8/src/inner.rs
+++ b/bb8/src/inner.rs
@@ -1,7 +1,6 @@
 use std::cmp::{max, min};
 use std::fmt;
 use std::future::Future;
-use std::sync::atomic::Ordering;
 use std::sync::{Arc, Weak};
 use std::time::{Duration, Instant};
 
@@ -86,7 +85,7 @@ where
     }
 
     pub(crate) async fn get(&self) -> Result<PooledConnection<'_, M>, RunError<M::Error>> {
-        let mut kind = StatsKind::Direct;
+        let mut get_kind = StatsKind::GetDirect;
         let mut wait_time_start = None;
 
         let future = async {
@@ -101,7 +100,7 @@ where
                     Some(conn) => PooledConnection::new(self, conn),
                     None => {
                         wait_time_start = Some(Instant::now());
-                        kind = StatsKind::Waited;
+                        get_kind = StatsKind::GetWaited;
                         self.inner.notify.notified().await;
                         continue;
                     }
@@ -116,8 +115,7 @@ where
                     Err(e) => {
                         self.inner
                             .statistics
-                            .connections_invalid_closed
-                            .fetch_add(1, Ordering::SeqCst);
+                            .record(StatsKind::ConnectionsInvalidClosed, 1);
                         self.inner.forward_error(e);
                         conn.state = ConnectionState::Invalid;
                         continue;
@@ -129,19 +127,18 @@ where
         let result = match timeout(self.inner.statics.connection_timeout, future).await {
             Ok(result) => result,
             _ => {
-                kind = StatsKind::TimedOut;
+                get_kind = StatsKind::GetTimedOut;
                 Err(RunError::TimedOut)
             }
         };
 
-        self.inner.statistics.record_get(kind);
+        self.inner.statistics.record(get_kind, 1);
 
         if let Some(wait_time_start) = wait_time_start {
             let wait_time = Instant::now() - wait_time_start;
             self.inner
                 .statistics
-                .get_waited_time_micro
-                .fetch_add(wait_time.as_micros() as u64, Ordering::SeqCst);
+                .record(StatsKind::GetWaitedTime, wait_time.as_micros() as u64);
         }
 
         result
@@ -167,8 +164,7 @@ where
                 if is_broken {
                     self.inner
                         .statistics
-                        .connections_broken_closed
-                        .fetch_add(1, Ordering::SeqCst);
+                        .record(StatsKind::ConnectionsBrokenClosed, 1);
                 }
                 let approvals = locked.dropped(1, &self.inner.statics);
                 self.spawn_replenishing_approvals(approvals);
@@ -214,8 +210,7 @@ where
                         .put(conn, Some(approval), self.inner.clone());
                     self.inner
                         .statistics
-                        .connections_created
-                        .fetch_add(1, Ordering::SeqCst);
+                        .record(StatsKind::ConnectionsCreated, 1);
                     return Ok(());
                 }
                 Err(e) => {
--- a/bb8/src/inner.rs
+++ b/bb8/src/inner.rs
@@ -1,6 +1,7 @@
 use std::cmp::{max, min};
 use std::fmt;
 use std::future::Future;
+use std::sync::atomic::Ordering;
 use std::sync::{Arc, Weak};
 use std::time::{Duration, Instant};
 
@@ -86,6 +87,7 @@ where
 
     pub(crate) async fn get(&self) -> Result<PooledConnection<'_, M>, RunError<M::Error>> {
         let mut kind = StatsKind::Direct;
+        let mut wait_time_start = None;
 
         let future = async {
             loop {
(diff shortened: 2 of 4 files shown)


bb8/src/inner.rs
```diff
@@ -111,6 +114,10 @@ where
                 match self.inner.manager.is_valid(&mut conn).await {
                     Ok(()) => return Ok(conn),
                     Err(e) => {
+                        self.inner
+                            .statistics
+                            .connections_invalid_closed
+                            .fetch_add(1, Ordering::SeqCst);
```
djc, line 120: We can just make this another `StatsKind`, right?

bb8/src/inner.rs
```diff
@@ -127,7 +134,16 @@ where
...
 
-        self.inner.statistics.record(kind);
+        self.inner.statistics.record_get(kind);
+
+        if let Some(wait_time_start) = wait_time_start {
+            let wait_time = Instant::now() - wait_time_start;
+            self.inner
+                .statistics
+                .get_waited_time_micro
+                .fetch_add(wait_time.as_micros() as u64, Ordering::SeqCst);
```
djc, line 144: Why micros vs millis or nanos?

bb8/src/internals.rs
```diff
@@ -139,22 +147,34 @@ where
         ApprovalIter { num: num as usize }
     }
 
-    pub(crate) fn reap(&mut self, config: &Builder<M>) -> ApprovalIter {
+    pub(crate) fn reap(&mut self, config: &Builder<M>) -> (ApprovalIter, u64, u64) {
+        let mut max_lifetime_closed: u64 = 0;
```
djc, line 151: Please avoid type annotations where possible.

bb8/src/internals.rs
```diff
@@ -255,10 +275,16 @@ pub(crate) struct AtomicStatistics {
     pub(crate) get_direct: AtomicU64,
     pub(crate) get_waited: AtomicU64,
     pub(crate) get_timed_out: AtomicU64,
+    pub(crate) get_waited_time_micro: AtomicU64,
+    pub(crate) connections_created: AtomicU64,
+    pub(crate) connections_broken_closed: AtomicU64,
+    pub(crate) connections_invalid_closed: AtomicU64,
+    pub(crate) connections_max_lifetime_closed: AtomicU64,
+    pub(crate) connections_max_idle_timeout_closed: AtomicU64,
```
djc, line 283: Let's skip the `connections_` prefixes.

bb8/src/internals.rs
```diff
@@ -255,10 +275,16 @@ pub(crate) struct AtomicStatistics {
...
 }
 
 impl AtomicStatistics {
-    pub(crate) fn record(&self, kind: StatsKind) {
+    pub(crate) fn record_get(&self, kind: StatsKind) {
```
djc, line 287: Given that we can use it for incrementing other (non-`get`) counters, prefer to revert this.

bb8/src/api.rs
```diff
@@ -99,6 +99,20 @@ pub struct Statistics {
     pub get_waited: u64,
     /// Total gets performed that timed out while waiting for a connection.
     pub get_timed_out: u64,
+    /// Total time accumulated waiting for a connection.
+    pub get_waited_time_micro: u64,
```
djc, line 103: Let's make this a `std::time::Duration`, and lose the `_micro` suffix.

bb8/src/inner.rs
pfreixes: jaja, good point. Ill try to do that!

bb8/src/inner.rs
pfreixes: I guess that micros would provide enough detail (nanos might be just noise) while we would keep the also the u64 large enough for storing many many many micros.

In any case since we move this to a Duration we will be storing behind the scenes IIRC micros using u128.

bb8/src/internals.rs
pfreixes: For all of them? at least at first sight the `connections_created`, `connections_broken_closed`, `connections_broken_closed` and `connections_invalid_closed` without the `connections_` prefix will seem a bit off.

Also it follows the same practice for the `get_` Kind, where we prefixed it and IMO helps to the user to understand the scope of them, same would happen for `connections_`

bb8/src/inner.rs
pfreixes: Oks now I remembered, everything I wanted to use an Atomic Type. I guess that we can try to use an `u128` internally and return a Duration (which IIRC is just a u128). If this works we could just forget about the discussion between nanos and micros.

pfreixes: Finally I went for using micros with u64 for converting later to Duration. If you want to use nanos, i can do that. But got the feeling that nanosecond resolution for this specific use case would not provide a lot of value (indeed micro most likely will not neither)

Ive just kept the type to `u64` just for having all atomics using the same type and helping me to keep the `record` interface simple.

If you want to to go for nanos and/or `u128` let me know.

pfreixes:
@djc this one is ready for a second review.

bb8/src/inner.rs
```diff
@@ -85,7 +85,8 @@ where
     }
 
     pub(crate) async fn get(&self) -> Result<PooledConnection<'_, M>, RunError<M::Error>> {
-        let mut kind = StatsKind::Direct;
+        let mut get_kind = StatsKind::GetDirect;
```
djc, line 88: Don't rename this `get_kind`, please.

bb8/src/inner.rs
```diff
@@ -122,12 +127,20 @@ where
...
 
-        self.inner.statistics.record(kind);
+        self.inner.statistics.record(get_kind, 1);
+
+        if let Some(wait_time_start) = wait_time_start {
+            let wait_time = Instant::now() - wait_time_start;
+            self.inner
+                .statistics
+                .record(StatsKind::GetWaitedTime, wait_time.as_micros() as u64);
+        }
```
djc, line 142: Let's wrap this up in a method on `AtomicStatistics` called `record_get()`.

bb8/src/internals.rs
```diff
@@ -49,7 +49,17 @@ where
...
-        locked.reap(&self.statics)
+        let (iter, max_idle_timeout_closed, max_lifetime_closed) = locked.reap(&self.statics);
+        drop(locked);
+
+        self.statistics
+            .record(StatsKind::ConnectionsMaxLifeTimeClosed, max_lifetime_closed);
+        self.statistics.record(
+            StatsKind::ConnectionsMaxIdleTimeoutClosed,
+            max_idle_timeout_closed,
+        );
```
djc, line 60: Let's wrap this in a separate method on `AtomicStatistics`, too, and keep all of the reaping-related changes in a separate commit or PR.

bb8/src/internals.rs
```diff
@@ -139,22 +149,34 @@ where
         ApprovalIter { num: num as usize }
     }
 
-    pub(crate) fn reap(&mut self, config: &Builder<M>) -> ApprovalIter {
+    pub(crate) fn reap(&mut self, config: &Builder<M>) -> (ApprovalIter, u64, u64) {
+        let mut max_lifetime_closed = 0;
+        let mut max_idle_timeout_closed = 0;
```
djc, line 154: Let's call these `closed_max_lifetime` and `closed_idle_timeout`.

bb8/src/internals.rs
djc: Okay, let's keep `connections_` but I'd like `closed` moved directly after `connections_` and rename `max_idle_timeout` to `idle_timeout`.

bb8/src/internals.rs
```diff
@@ -273,12 +319,30 @@ impl From<&AtomicStatistics> for Statistics {
...
+                .load(Ordering::SeqCst),
         }
     }
 }
 
 pub(crate) enum StatsKind {
-    Direct,
-    Waited,
-    TimedOut,
+    GetDirect,
```
djc, line 339: I think we'll want to split this into `StatsGet { Waited, Direct, TimedOut }`.

djc (commented):
Okay, I want a clean commit history with separate logical changes, so this should end up being in a different commit. I usually handle this by using interactive rebases but feel free to use separate PRs if that's simpler for you. I think we need the following changes:

* Rename `StatsKind` to `StatsGet`
* Rename `record()` to `record_get()`
* Track `wait_time`: add `wait_time` variable in `get()` and change `record_get()` to accept the `Option<Duration>`, should end up in `get_wait_time_micros` in `AtomicStatistics` and in `get_wait_time` in `Statistics`
* Track `closed_broken` and `closed_created`: recreate `AtomicStatistics::record()` and `StatsKind` for these
* Track `closed_max_lifetime` and `closed_idle_timeout`: create a separate method `AtomicStatistics::reaped()` for these

bb8/src/internals.rs
pfreixes: @djc just for having an agreement on the new kinds and record operations that we would have within `AtomicStatistics`, we can go for this

```rust
pub(crate) enum StatsGet {
    Waited,
    Direct,
    TimedOut,
}

pub(crate) enum StatsConnectionClosed {
    Broken,
    Invalid,
}
```

Note: we should not need more stats kind since the used `record_*` methods for stats like the ones coming from the reap function will just call an ad-hoc function like `record_connections_reaped`

And then we would have the following `record_*` methods within the `AtomicStatistics`:

- `record_get` (takes the `StatsGet` parameter)
- `record_connection_closed`  (takes the `StatsConnectionClosed` parameter)
- `record_connections_reaped`
- `record_get_wait_time` 
- `record_connection_created`

Let me know if this works for you and i can create a new PR with separated commits for each group, starting with the rename of the `record` to `record_get`

djc: No, as suggested in my previous message we should reuse a simple `record()` with `StatsKind` for both `connection_closed` and `connection_created`, and `record_get()` should also handle `get_wait_time`.

pfreixes: Oks, let me provide a new draft - sorry for the back and forth here :

```
pub(crate) enum StatsGet {
    Waited,
    Direct,
    TimedOut,
    WaitTime
}

pub(crate) enum StatsConnectionClosed {
    Broken,
    Invalid,
}
```

And following `record` methods

- `record` which would take a `StatsConnectionClosed` kind type
- `record_get` for recording any stats related to get including wait time
- `record_connections_reaped` for recording the reaped connections either idle timeout or max lifetime

How does it sound? is this aligned with your expectations?

djc: No. Let's do one PR at a time, please implement the first chunk and I'll give feedback. Please closely read my feedback from this PR because you've still missed some of it.

PR #205, opened 2024-06-12
Adds new statistics attributes for tracking connections closed

pfreixes:
The two new attributes `connections_closed_broken` and `connections_closed_invalid` can be used for respectively understand how many connections were closed due to be considered broken or invalid.

The `StatsKind` internal enum was also renamed to `StatsGetKind` for better clarity.

Note: A new PR for tracking the reaped connections will be created on top of these changes.

### DIFF
--- a/bb8/src/internals.rs
+++ b/bb8/src/internals.rs
@@ -259,11 +259,11 @@ pub(crate) struct AtomicStatistics {
 }
 
 impl AtomicStatistics {
-    pub(crate) fn record_get(&self, kind: StatsKind, wait_time_start: Option<Instant>) {
+    pub(crate) fn record_get(&self, kind: StatsGetKind, wait_time_start: Option<Instant>) {
         match kind {
-            StatsKind::Direct => self.get_direct.fetch_add(1, Ordering::SeqCst),
-            StatsKind::Waited => self.get_waited.fetch_add(1, Ordering::SeqCst),
-            StatsKind::TimedOut => self.get_timed_out.fetch_add(1, Ordering::SeqCst),
+            StatsGetKind::Direct => self.get_direct.fetch_add(1, Ordering::SeqCst),
+            StatsGetKind::Waited => self.get_waited.fetch_add(1, Ordering::SeqCst),
+            StatsGetKind::TimedOut => self.get_timed_out.fetch_add(1, Ordering::SeqCst),
         };
 
         if let Some(wait_time_start) = wait_time_start {
@@ -285,7 +285,7 @@ impl From<&AtomicStatistics> for Statistics {
     }
 }
 
-pub(crate) enum StatsKind {
+pub(crate) enum StatsGetKind {
     Direct,
     Waited,
     TimedOut,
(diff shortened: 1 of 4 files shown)


bb8/src/internals.rs
```diff
@@ -290,3 +305,8 @@ pub(crate) enum StatsGetKind {
     Waited,
     TimedOut,
 }
+
+pub(crate) enum StatsConnectionClosedKind {
```
djc, line 309: Since there's nothing "special" about these values, let's just call this `StatsKind` again.

bb8/src/internals.rs
```diff
@@ -272,6 +274,17 @@ impl AtomicStatistics {
                 .fetch_add(wait_time.as_micros() as u64, Ordering::SeqCst);
         }
     }
+
+    pub(crate) fn record(&self, kind: StatsConnectionClosedKind) {
+        match kind {
+            StatsConnectionClosedKind::Broken => self
```
djc, line 280: Maybe yield `&self.connections_closed_broken` from the match arm and then move the `.fetch_add(1, Ordering::SeqCst)` call out of the `match?

djc (commented):
Please squash the suggested changes into your second commit.

bb8/src/internals.rs
pfreixes: TBH i would prefer to keep the ConnectionClosed since this is used for recording closed connections, but Ill make the change

bb8/src/internals.rs
pfreixes: I can try it, but to me seems to overcomplicate things for just trying to make the code less verbose but making it less readable.

pfreixes:
@djc squashed requested changes into my latest commit PTAL one you have a moment

bb8/src/internals.rs
djc: It can be used for any kind of simple increment of `AtomicU64`.

bb8/src/internals.rs
djc: I prefer the conciseness and reduced duplication.

bb8/src/internals.rs
pfreixes: Its ok, lets see how `record` evolves .... BTW added the changes that you asked

djc:
I'm not seeing the changes I suggested appear in your commit, I think? Are you sure your squashing worked as expected?

bb8/src/internals.rs
djc: If we rename the enum might also rename the variants to be like `ClosedBroken`, `ClosedInvalid`.

pfreixes:
@djc, forgot to add the files changed 🤦

bb8/src/internals.rs
```diff
@@ -272,6 +274,14 @@ impl AtomicStatistics {
                 .fetch_add(wait_time.as_micros() as u64, Ordering::SeqCst);
         }
     }
+
+    pub(crate) fn record(&self, kind: StatsKind) {
+        let stats = match kind {
+            StatsKind::ClosedBroken => &self.connections_closed_broken,
+            StatsKind::ClosedInvalid => &self.connections_closed_invalid,
+        };
+        stats.fetch_add(1, Ordering::SeqCst);
```
djc, line 283: Sorry, I meant you can chain the `fetch_add()` directly onto the `match { .. }` without binding it.

djc (commented):
One minor nit!

djc (approved):

PR #206, opened 2024-06-13
Adds new statistics attributes for reaped connections

pfreixes:
The two new statistics, `connections_closed_max_lifetime` and `connections_closed_idle_timeout` the total number of connections that were reaped due to reching the max lifetime or the idle timeout.


djc (approved):

djc:
Thanks!

PR #208, opened 2024-06-14
Bump to 0.8.4 version

pfreixes:
We need to have [this](https://github.com/djc/bb8/pull/207) merged first

### DIFF
--- a/bb8/Cargo.toml
+++ b/bb8/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "bb8"
-version = "0.8.3"
+version = "0.8.4"
 edition = "2021"
 rust-version = "1.63"
 description = "Full-featured async (tokio-based) connection pool (like r2d2)"


bb8/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "bb8"
-version = "0.8.3"
+version = "0.9.0"
```
djc, line 3: Why not 0.8.4? I think all of your changes have been semver-compatible?

pfreixes: Since its adding a new feature and not patching anything got the feeling that minor bump would be the most suitable one to bump.

pfreixes: And we keep compatibility

djc: That's not how semver works in Cargo: 0.x to 0.y is a semver-incompatible bump, there is no distinction between feature and fix versions in 0.x versions.

pfreixes: I was not aware about this change from Cargo compared to what we have historically used major, minor and patched versions.

pfreixes: Ill put the 0.8.4

pfreixes: Done, and thanks for the insights here about how Cargo works with versions

djc (approved):

djc:
* Published bb8 v0.8.4 at registry `crates-io`
* [new tag]         v0.8.4 -> v0.8.4

Thanks!

pfreixes:
Seems that the documentation is not providing all the details https://docs.rs/bb8/0.8.4/bb8/struct.State.html did we miss something for having the new `Statistic` struct being added in the documentation?

djc:
Ahh, we probably need to re-export `Statistics` in `lib.rs`. Do you want to send a PR to do that, including another commit to bump the version to 0.8.5? (You can run `cargo doc --open --no-deps` to check the generated docs before submitting the PR.)

pfreixes:
Yeps, give me some seconds

pfreixes:
Here we go https://github.com/djc/bb8/pull/209

PR #210, opened 2024-06-20
Use Relaxed ordering mode

pfreixes:
The Relaxed ordering mode is much more efficient than SeqCst and its good enough for the use case of either loading statistics or updating the current statistic with a delta.

I had a wrong understanding of the ordering mode and I apologize for the confusion. I have updated the code to use Relaxed ordering mode instead of SeqCst. This will make the code more efficient and will be good enough for the use case of either loading statistics or updating the current statistic with a delta.

As an example here for example what ordering mode is used by tokio for [incrementing](https://github.com/tokio-rs/tokio/blob/9a75d6f7f70fd81f589a8209c548944bc3d05f84/tokio/src/util/metric_atomics.rs#L75) the number of threads created.

There is a good explanation about the semantics of the ordering parameter in this book, specifically this [section](https://marabos.nl/atomics/memory-ordering.html#happens-before), which explains that the memory ordering is about the ordering of the atomic instructions rather than the synchronization of the threads for a specific memory address

### DIFF
--- a/bb8/src/internals.rs
+++ b/bb8/src/internals.rs
@@ -282,15 +282,15 @@ pub(crate) struct AtomicStatistics {
 impl AtomicStatistics {
     pub(crate) fn record_get(&self, kind: StatsGetKind, wait_time_start: Option<Instant>) {
         match kind {
-            StatsGetKind::Direct => self.get_direct.fetch_add(1, Ordering::SeqCst),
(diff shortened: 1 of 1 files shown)


djc (approved):

djc:
Makes sense, thanks!

PR #214, opened 2024-07-27
Add `Pool::add`

### LINKED ISSUE #212: [feature] Add/Get New Pooled Connections
I have a use case where I want to periodically re-evaluate my login credentials against the underlying database.

Using something like `Pool::get()` doesn't work, because it doesn't necessarily mean that the connection is new - it could be from the pool instead. Luckily, this is pretty trivial today with `Pool::dedicated_connection()`:

```rust
let pool = Pool::builder().build(mgr).await.unwrap();
let mut interval = time::interval(time::Duration::from_secs(5));
loop {
  tokio::select! {
    _ = interval.tick() => match pool.dedicated_connection().await {
      Ok(_) => (),
      Err(err) => handle_connection_error(err),
  }
}
```

However, establishing a connection against my database is expensive, and I don't want to just throw away the connection after its established. Instead, it would be great if that new connection could be added to the pool.

I see two obvious ways of going about this:

1. A method to _add_ a connection to the pool:

```rust
/// Adds a connection to the pool.
pub fn add(&self, conn: M::Connection) -> ()
```

```rust
match pool.dedicated_connection().await {
  Ok(conn) => pool.add(conn),
  Err(err) => handle_connection_error(err),
}
```

2. A method to get a _new_ connection from the pool:

```rust
/// Retrieves a new connection from the pool.
pub async fn get_new(&self) -> Result<PooledConnection<'_, M>, RunError<M::Error>>
```

```rust
match pool.get_new().await {
  Ok(_pooled_connection) => (), // returned to pool on drop
  Err(err) => handle_connection_error(err),
}
```

I'm leaning towards the first option, since it's both easier to implement and more flexible. For example, you can create connections independent from the pool and add them to be managed later.

I would be happy to write the code for this feature if it's something you're ok with adding to the project.

tneely:
Fixes #212

This adds `Pool::add`, which allows for externally created connections to be added and managed by the pool. If the pool is at maximum capacity when this method is called, it will return the input connection as part of the Err response.

I considered allowing `Pool:add` to ignore `max_size` when adding to the pool, but felt it could lead to confusion if the pool is allowed to exceed its capacity in this specific case.

This change required making PoolInternals::approvals visible within the crate to get the approval needed to add a new connection. The alternative would have required defining a new pub(crate) method for this specific use case, which feels worse. I'm open to suggestions on how to more cleanly integrate this change into the package.

### DIFF
--- a/bb8/src/api.rs
+++ b/bb8/src/api.rs
@@ -75,6 +75,14 @@ impl<M: ManageConnection> Pool<M> {
     pub fn state(&self) -> State {
         self.inner.state()
     }
+
+    /// Adds a connection to the pool.
+    ///
+    /// If the connection is broken, or the pool is at capacity, the
+    /// connection is not added and instead returned to the caller in Err.
+    pub fn add(&self, conn: M::Connection) -> Result<(), AddError<M::Connection>> {
+        self.inner.try_put(conn)
+    }
 }
 
 /// Information about the state of a `Pool`.
@@ -523,6 +531,33 @@ where
 {
     fn from(error: E) -> Self {
         Self::User(error)
+    }
+}
+
+/// Error type returned by `Pool::add(conn)`
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum AddError<C> {
+    /// The connection was broken before it could be added.
+    Broken(C),
+    /// Unable to add the connection to the pool due to insufficient capacity.
+    NoCapacity(C),
+}
+
+impl<E: error::Error + 'static> fmt::Display for AddError<E> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            AddError::Broken(_) => write!(f, "The connection was broken before it could be added"),
+            AddError::NoCapacity(_) => write!(
+                f,
+                "Unable to add the connection to the pool due to insufficient capacity"
+            ),
+        }
+    }
+}
+
+impl<E: error::Error + 'static> error::Error for AddError<E> {
+    fn source(&self) -> Option<&(dyn error::Error + 'static)> {
+        None
     }
 }
 
--- a/bb8/src/inner.rs
+++ b/bb8/src/inner.rs
@@ -9,7 +9,9 @@ use futures_util::TryFutureExt;
 use tokio::spawn;
 use tokio::time::{interval_at, sleep, timeout, Interval};
 
-use crate::api::{Builder, ConnectionState, ManageConnection, PooledConnection, RunError, State};
+use crate::api::{
+    AddError, Builder, ConnectionState, ManageConnection, PooledConnection, RunError, State,
+};
 use crate::internals::{Approval, ApprovalIter, Conn, SharedPool, StatsGetKind, StatsKind};
 
 pub(crate) struct PoolInner<M>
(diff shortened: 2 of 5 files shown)


bb8/src/inner.rs
```diff
@@ -161,6 +161,20 @@ where
         }
     }
 
+    /// Adds an external connection to the pool if there is capacity for it.
+    pub(crate) fn try_put(&self, conn: M::Connection) -> Result<(), M::Connection> {
```
djc, line 165: It might be better to move this logic into a method on `SharedPool`, which would also allow us to keep `approvals()` private. Then, I think we want a layer here which would run `manager.has_broken()` on the new `conn` before we add it. Might need another error enum to accomodate the different error paths.

bb8/src/inner.rs
```diff
@@ -161,6 +161,20 @@ where
...
+        let mut locked = self.inner.internals.lock();
+        let mut approvals = locked.approvals(&self.inner.statics, 1);
+        match approvals.next() {
```
djc, line 168: Suggest spelling this as `let Some(approval) = approvals.next() else { return Err(conn); }`.

bb8/src/inner.rs
tneely: That makes sense! Let me move things around

bb8/src/inner.rs
tneely: Sounds good!

tneely:
Sorry for the delay - I lost track of this PR at some point. I've moved a few things around based on your feedback and also added an `AddError` enum to differentiate between when the connection is broken and when there's no capacity.

bb8/src/api.rs
```diff
@@ -547,3 +555,36 @@ impl<E> ErrorSink<E> for NopErrorSink {
...
+pub enum AddError<C> {
+    /// The connection was broken before it could be added.
+    Broken(C),
+    /// Unable to add the connection to the pool due to insufficient capacity.
+    NoCapacity(C),
+}
+
+impl<E> fmt::Display for AddError<E>
+where
+    E: error::Error + 'static,
```
djc, line 570: Nit: please put these bounds into the `impl` declaration for conciseness (so `impl<E: error::Error + 'static>`).

bb8/src/api.rs
```diff
@@ -547,3 +555,36 @@ impl<E> ErrorSink<E> for NopErrorSink {
         Box::new(*self)
     }
 }
+
+/// Error type returned by `Pool::add(conn)`
+#[derive(Debug, Clone, PartialEq, Eq)]
```
djc, line 560: Nit: let's move this next to `RunError` (so above `ErrorSink`).

djc (commented):
Looking pretty good, some small suggestions left!

bb8/tests/test.rs
```diff
@@ -104,6 +104,37 @@ where
...
+    _c: PhantomData<C>,
+}
+
+impl<C> BrokenConnectionManager<C> {
+    fn new() -> Self {
+        BrokenConnectionManager { _c: PhantomData }
+    }
+}
+
+#[async_trait]
```
djc, line 117: Sorry, missed this previously: I want to review this move separately, so please isolate it in a separate commit (and squash all the other commits) together.

djc (approved):
This mostly looks good, needs one tweak to the commit history.

bb8/tests/test.rs
djc: The idea here is that I want the move of the existing code here isolated in a separate commit from adding the new test -- sorry if that wasn't clear from my previous comment.

tneely: I think I understand? Let me try again.

djc: Not quite, but I've taken the liberty of fixing it up.

djc:
Thanks for your patience through a number of iterations!

tneely:
You're welcome, thanks for working through the PR with me!

PR #217, opened 2024-08-19
testcontainers instead of manual docker run

awnion:
What do you think if I make it for all examples?

Main benefit: e.g. for postgres port can be anything and will be managed by testcontainers while use can refer it like default `5432`

```rust
    let db_port = postgres_container.get_host_port_ipv4(5432).await.unwrap();
```


awnion:
To be honest just played with test containers and it happened to be your repo example. 
If you think it's not motivated -- fair enough. I do not insist :)

PR #223, opened 2024-10-14
Fix leaky connections

### LINKED ISSUE #221: `pool.get()` can acquire more connections than are necessary
### Issue

It's possible to trigger more approvals than are necessary, in turn grabbing more connections than we need. This happens when we drop a connection. The drop produces a notify, which doesn't get used until the pool is empty. The first `Pool::get()` call on an empty pool will spawn an connect task, immediately complete `notify.notified().await`, then spawn a second connect task. Both will connect and we'll end up with 1 more connection than we need.

This gets worse if one of those connections is broken when it gets returned to the pool. We end up calling `notify.notify_waiters();`, which means all connections in the `PoolInner::get()` loop could spawn an extra connect task. This should just be a single `notify.notify_one();` call.

https://github.com/djc/bb8/blob/cb996975a2cdf1a7d73ac45b9b8602d690bb6ef9/bb8/src/inner.rs#L161

### Affected Versions

I have only tested on 0.8.5

### Fix

Not sure on this one. I think removing the `notify.notify_one()` call from the drop path would fix this specific issue, but it would hurt performance and could cause other issues (e.g. if the pool is at max capacity and we have no inflight connect tasks). 

We should definitely change `notify.notify_waiters()` -> `notify.notify_one()` in the broken connection case if possible.

### Reproduction

```rust
#[tokio::test]
async fn test_notify() {
    let pool = Pool::builder()
        .min_idle(0)
        .max_size(100)
        .queue_strategy(QueueStrategy::Lifo)
        .build(OkManager::<FakeConnection>::new())
        .await
        .unwrap();

    // The first get should
    // 1) see nothing in the pool,
    // 2) spawn a single connect task,
    // 3) get notified of the new connection and grab it from the pool
    let conn_0 = pool.get().await.expect("should connect");
    // Dropping the connection queues up a notify
    drop(conn_0);
    // The second get should
    // 1) see the first connection in the pool and grab it
    // NB: This does not consume the notify
    let _conn_1: PooledConnection<OkManager<FakeConnection>> = pool.get().await.expect("should connect");
    // The third get will
    // 1) see nothing in the pool,
    // 2) spawn a single connect task,
    // 3) get notified from the dropped connection
    // 4) see nothing in the pool,
    // 5) spawn a single connect task (oh no!),
    // 6) get notified of the new connection and grab it from the pool
    let _conn_2: PooledConnection<OkManager<FakeConnection>> = pool.get().await.expect("should connect");

    assert_eq!(pool.state().connections, 2);
}
```

```
thread 'test_notify' panicked at bb8/tests/test.rs:1160:5:
assertion `left == right` failed
  left: 3
 right: 2
```

tneely:
Fixes #221

It's possible to trigger more approvals than are necessary, in turn
grabbing more connections than we need. This happens when we drop a
connection. The drop produces a notify, which doesn't get used until the
pool is empty. The first `Pool::get()` call on an empty pool will spawn
an connect task, immediately complete `notify.notified().await`, then
spawn a second connect task. Both will connect and we'll end up with 1
more connection than we need.

Rather than address the notify issue directly, this fix introduces some
bookkeeping that tracks the number of open `pool.get()` requests we have
waiting on connections. If the number of pending connections >= the
number of pending gets, we will not spawn any additional connect tasks.

~~I have additionally changed `notify.notify_waiters();` to a
`notify.notify_one();` call in the broken connection branch. A single
broken connection should only need to notify one pending task to spawn a
new connect task, not all of them.~~ See https://github.com/djc/bb8/pull/225

### DIFF
--- a/bb8/src/inner.rs
+++ b/bb8/src/inner.rs
@@ -91,6 +91,7 @@ where
         let mut wait_time_start = None;
 
         let future = async {
+            let _guard = self.inner.request();
             loop {
                 let (conn, approvals) = self.inner.pop();
                 self.spawn_replenishing_approvals(approvals);
--- a/bb8/src/internals.rs
+++ b/bb8/src/internals.rs
@@ -22,6 +22,27 @@ where
     pub(crate) statistics: AtomicStatistics,
 }
 
+pub(crate) struct GetGuard<M: ManageConnection + Send> {
+    inner: Arc<SharedPool<M>>,
+}
+
+impl<M: ManageConnection + Send> GetGuard<M> {
+    fn new(inner: Arc<SharedPool<M>>) -> Self {
+        {
+            let mut locked = inner.internals.lock();
+            locked.in_flight += 1;
+        }
+        GetGuard { inner }
+    }
+}
+
+impl<M: ManageConnection + Send> Drop for GetGuard<M> {
+    fn drop(&mut self) {
+        let mut locked = self.inner.internals.lock();
+        locked.in_flight -= 1;
+    }
+}
+
 impl<M> SharedPool<M>
 where
     M: ManageConnection + Send,
@@ -41,10 +62,20 @@ where
         let conn = locked.conns.pop_front().map(|idle| idle.conn);
         let approvals = match &conn {
             Some(_) => locked.wanted(&self.statics),
-            None => locked.approvals(&self.statics, 1),
+            None => {
+                let approvals = match locked.in_flight > locked.pending_conns {
+                    true => 1,
+                    false => 0,
+                };
+                locked.approvals(&self.statics, approvals)
+            }
         };
 
         (conn, approvals)
+    }
+
+    pub(crate) fn request(self: &Arc<Self>) -> GetGuard<M> {
+        GetGuard::new(self.clone())
     }
 
     pub(crate) fn try_put(self: &Arc<Self>, conn: M::Connection) -> Result<(), M::Connection> {
@@ -81,6 +112,7 @@ where
     conns: VecDeque<IdleConn<M::Connection>>,
     num_conns: u32,
     pending_conns: u32,
+    in_flight: u32,
 }
 
 impl<M> PoolInternals<M>
@@ -202,6 +234,7 @@ where
             conns: VecDeque::new(),
             num_conns: 0,
             pending_conns: 0,
+            in_flight: 0,
         }
     }
 }
--- a/bb8/tests/test.rs
+++ b/bb8/tests/test.rs
@@ -1068,3 +1068,37 @@ async fn test_add_checks_broken_connections() {
     let res = pool.add(conn);
     assert!(matches!(res, Err(AddError::Broken(_))));
 }
+
+#[tokio::test]
+async fn test_reuse_on_drop() {
+    let pool = Pool::builder()
+        .min_idle(0)
+        .max_size(100)
+        .queue_strategy(QueueStrategy::Lifo)
+        .build(OkManager::<FakeConnection>::new())
+        .await
+        .unwrap();
+
+    // The first get should
+    // 1) see nothing in the pool,
+    // 2) spawn a single replenishing approval,
+    // 3) get notified of the new connection and grab it from the pool
+    let conn_0 = pool.get().await.expect("should connect");
+    // Dropping the connection queues up a notify
+    drop(conn_0);
+    // The second get should
+    // 1) see the first connection in the pool and grab it
+    let _conn_1: PooledConnection<OkManager<FakeConnection>> =
+        pool.get().await.expect("should connect");
+    // The third get will
+    // 1) see nothing in the pool,
+    // 2) spawn a single replenishing approval,
+    // 3) get notified of the new connection,
+    // 4) see nothing in the pool,
+    // 5) _not_ spawn a single replenishing approval,
+    // 6) get notified of the new connection and grab it from the pool
+    let _conn_2: PooledConnection<OkManager<FakeConnection>> =
+        pool.get().await.expect("should connect");
+
+    assert_eq!(pool.state().connections, 2);
+}


djc:
Thanks for your work on this!

bb8/src/inner.rs
```diff
@@ -158,7 +159,7 @@ where
                 }
                 let approvals = locked.dropped(1, &self.inner.statics);
                 self.spawn_replenishing_approvals(approvals);
-                self.inner.notify.notify_waiters();
+                self.inner.notify.notify_one();
```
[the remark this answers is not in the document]
tneely: Nothing broke when I changed it independently. I can move this into its own commit

bb8/src/internals.rs
```diff
@@ -81,6 +109,7 @@ where
     conns: VecDeque<IdleConn<M::Connection>>,
     num_conns: u32,
     pending_conns: u32,
+    inflight_gets: u32,
```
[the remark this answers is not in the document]
tneely: Hmm I don't think it's possible, unless there's other state I'm not aware of.

In the following table I'm modeling the internal state following 3 gets, a drop + get, and a notify from the drop. During this time, none of the pending connect tasks have returned with a connection.

| Time | Action | Conns | Idle | Pending | In-Flight |
|--------|--------|--------|--------|--------|--------|
| 0 | - | 10 | 0 | 0 | 0 |
| 1 | Get | 10 | 0 | 1 | 1 |
| 2 | Get | 10 | 0 | 2 | 2 |
| 3 | Get | 10 | 0 | 3 | 3 |
| 4 | Drop | 10 | 1 | 3 | 3 |
| 5 | Get | 10 | 0 | 3 | 3 |
| 6 | Get_Notify | 10 | 0 | 3 | 3 |

In this case, a 4th `Get` is able to cut in front of the in-flight `Get`s, immediately taking the idle connection. When the `Get` at `t1` is notified, it sees nothing in the pool and 3 pending connect tasks. However in isolation it doesn't know that one of those 3 pending are for it, or that it needs to spawn its own connect task.

`num_conns` isn't particularly helpful here since `Get` only cares about what's available in the idle pool. And `pending_conns` by itself isn't enough, since we don't want 3 in-flight `Get`s to be waiting on a single pending connection.

bb8/src/inner.rs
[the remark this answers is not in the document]
tneely: I pulled this into its own PR just to make it clear the tests still pass - https://github.com/djc/bb8/pull/225. It's not necessary for this fix since it still holds even if we bounce all the waiters.

bb8/src/internals.rs
```diff
@@ -41,12 +62,19 @@ where
...
+                locked.approvals(&self.statics, approvals)
+            }
         };
 
         (conn, approvals)
     }
 
+    pub(crate) fn request(self: &Arc<Self>) -> GetGuard<M> {
+        GetGuard::new(self.clone())
+    }
```
djc, line 76: It feels like `request()` and `pop()` should be merged into a single method, which I'm thinking should yield a custom enum, maybe like this:

```rust
enum Requested {
    Connection(Conn<M::Connection>),
    Waiting(GetGuard<M>),
}
```

bb8/src/internals.rs
```diff
@@ -22,6 +22,27 @@ where
...
+pub(crate) struct GetGuard<M: ManageConnection + Send> {
+    inner: Arc<SharedPool<M>>,
+}
+
+impl<M: ManageConnection + Send> GetGuard<M> {
+    fn new(inner: Arc<SharedPool<M>>) -> Self {
+        {
+            let mut locked = inner.internals.lock();
+            locked.inflight_gets += 1;
+        }
```
djc, line 34: Nit: there's not really any point in this scope, right?

bb8/src/internals.rs
```diff
@@ -41,12 +62,19 @@ where
         let conn = locked.conns.pop_front().map(|idle| idle.conn);
         let approvals = match &conn {
             Some(_) => locked.wanted(&self.statics),
-            None => locked.approvals(&self.statics, 1),
+            None => {
+                let approvals = min(1, locked.inflight_gets.saturating_sub(locked.pending_conns));
+                locked.approvals(&self.statics, approvals)
```
djc, line 67: I don't find this logic very clear. Maybe something like this?

```rust
let approvals = match locked.inflight_gets < locked.pending_conns {
     true => 1,
     false => 0,
};
```

(Hope I didn't get the condition inverted there.)

bb8/src/internals.rs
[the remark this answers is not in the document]
djc: Okay, makes sense.

bb8/src/internals.rs
```diff
@@ -202,6 +231,7 @@ where
             conns: VecDeque::new(),
             num_conns: 0,
             pending_conns: 0,
+            inflight_gets: 0,
```
djc, line 234: Would prefer to rename this to `in_flight`.

bb8/src/internals.rs
tneely: This approach makes sense. It keeps tracking responsibilities more separated between inner and internals.

bb8/src/internals.rs
tneely: Fair, I can remove it

bb8/src/internals.rs
tneely: Can do!

bb8/src/internals.rs
tneely: Yeah that definitely reads better, thanks!

bb8/src/internals.rs
tneely: Although if we put this in `pop()` we'd get the guard inside the `loop`, which means we'd lose it on the next iteration

bb8/src/internals.rs
tneely: Ah sorry, I forgot I had this because I need to drop the MutexGuard before I can create the GetGuard.

```
cannot move out of `inner` because it is borrowed
```

Adding the scope felt better than doing a `.clone()`

djc:
See #226.

PR #224, opened 2024-10-15
Reap expired connections on drop

tneely:
The reaper only runs against the connections in its idle pool. This is fine for reaping idle connections, but for hotly contested connections beyond their maximum lifetime this can prove problematic.

Consider an active connection beyond its lifetime and a reaper that runs every 3 seconds:
- [t0] Connection is idle
- [t1] Connection is active
- [t2] Reaper runs, does not see connection
- [t3] Connection is idle

This pattern can repeat infinitely with the connection never being reaped.

By checking the max lifetime on drop, we can ensure that expired connections are reaped in a reason amount of time (assuming they eventually do get dropped).

### DIFF
--- a/bb8/src/inner.rs
+++ b/bb8/src/inner.rs
@@ -149,12 +149,18 @@ where
             "handled in caller"
         );
 
+        let max_lifetime = self.inner.statics.max_lifetime;
+        let is_expired = max_lifetime.map_or(false, |lt| conn.is_expired(Instant::now() - lt));
+        let is_broken = self.inner.manager.has_broken(&mut conn.conn);
+
         let mut locked = self.inner.internals.lock();
-        match (state, self.inner.manager.has_broken(&mut conn.conn)) {
+        match (state, is_broken || is_expired) {
             (ConnectionState::Present, false) => locked.put(conn, None, self.inner.clone()),
-            (_, is_broken) => {
+            _ => {
                 if is_broken {
                     self.inner.statistics.record(StatsKind::ClosedBroken);
+                } else if is_expired {
+                    self.inner.statistics.record_connections_reaped(0, 1);
                 }
                 let approvals = locked.dropped(1, &self.inner.statics);
                 self.spawn_replenishing_approvals(approvals);
--- a/bb8/src/internals.rs
+++ b/bb8/src/internals.rs
@@ -169,7 +169,7 @@ where
                 }
             }
             if let Some(lifetime) = config.max_lifetime {
-                if now - conn.conn.birth >= lifetime {
+                if conn.conn.is_expired(now - lifetime) {
                     closed_max_lifetime += 1;
                     keep &= false;
                 }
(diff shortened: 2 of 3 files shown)


bb8/tests/test.rs
```diff
@@ -585,6 +585,40 @@ async fn test_max_lifetime() {
...
+        .connection_timeout(Duration::from_secs(1))
+        .reaper_rate(Duration::from_secs(999))
+        .build(manager)
+        .await
+        .unwrap();
+
+    let conn = pool.get().await;
+
+    // And wait.
+    tokio::time::sleep(Duration::from_secs(2)).await;
```
djc, line 613: Nit: prefer to just import `sleep()` for this.

bb8/src/internals.rs
```diff
@@ -169,7 +169,7 @@ where
                 }
             }
             if let Some(lifetime) = config.max_lifetime {
-                if now - conn.conn.birth >= lifetime {
+                if conn.conn.is_expired(now - lifetime) {
```
djc, line 172: Nit: would like to see this split out into a separate commit before the main change.

bb8/src/inner.rs
```diff
@@ -149,12 +149,19 @@ where
             "handled in caller"
         );
 
+        let max_lifetime = self.inner.statics.max_lifetime;
+        let is_expired = max_lifetime.map_or(false, |lt| conn.is_expired(Instant::now() - lt));
```
djc, line 153: Suggest we bind `let is_broken = self.inner.manager.has_broken(&mut conn.conn);` here and then use `is_broken || is_expired` in the match scrutinee.

djc (commented):
Thanks, this makes sense. Suggest some minor changes.

Can you talk about what you're using bb8 for? Would be good to know.

tneely:
I'm using bb8 inside of [PgCat](https://github.com/postgresml/pgcat) to do transaction pooling. The servers I'm connecting to get cycled periodically for security reasons, so if the reaper lags too far behind then we end up querying a dead server and reporting the connection as broken. I'd rather reap it before we get to that point.

bb8/src/internals.rs
tneely: Can do!

bb8/src/inner.rs
```diff
@@ -149,12 +149,20 @@ where
...
+        let is_broken = self.inner.manager.has_broken(&mut conn.conn);
+
         let mut locked = self.inner.internals.lock();
-        match (state, self.inner.manager.has_broken(&mut conn.conn)) {
-            (ConnectionState::Present, false) => locked.put(conn, None, self.inner.clone()),
-            (_, is_broken) => {
+        match (state, is_broken || is_expired) {
+            (ConnectionState::Present, false) if !is_expired => {
```
djc, line 158: So we don't need `if !is_expired` anymore, right?

tneely: Good catch, I totally missed that

djc:
See follow-up in #226.

PR #226, opened 2024-10-17
Polish for fixes from #223 and #224

djc:
@tneely thanks for your efforts so far! I have some nits to pick with your PRs but figured it might be more expedient (especially given the ~24-hour RTT) if I just took care of them myself. Will retain your authorship and make minor changes.


tneely:
Thanks for getting this over the finish line!

PR #233, opened 2025-01-07
Use portable-atomic when AtomicU64 is not available.

### LINKED ISSUE #232: AtomicU64 unresolved symbol
`AtomicU64` does not exist in some systems.

I was trying to compile sccache in a powerpc (32-bit) vm with ArchPOWER.
It failed because of this library:

```
...
   Compiling bb8 v0.8.6
error[E0432]: unresolved import `std::sync::atomic::AtomicU64`
 --> /home/user/.cargo/registry/src/index.crates.io-9850fa1df3136423/bb8-0.8.6/src/internals.rs:3:25
  |
3 | use std::sync::atomic::{AtomicU64, Ordering};
  |                         ^^^^^^^^^
  |                         |
  |                         no `AtomicU64` in `sync::atomic`
  |                         help: a similar name exists in the module: `AtomicU32`

For more information about this error, try `rustc --explain E0432`.
error: could not compile `bb8` (lib) due to 1 previous error
error: failed to compile `sccache v0.9.0`, intermediate artifacts can be found at `/tmp/cargo-installkWejfU`.
...
```

flaviojs:
*edit* Only fixes #232 if applied to a 0.8.x branch.

AtomicU64 was introduced in v0.8.4
The current version is 0.9.0, please update 0.8.x too.

### DIFF
--- a/bb8/src/internals.rs
+++ b/bb8/src/internals.rs
@@ -1,6 +1,10 @@
+#[cfg(not(target_has_atomic = "64"))]
+use portable_atomic::AtomicU64;
 use std::cmp::min;
 use std::collections::VecDeque;
-use std::sync::atomic::{AtomicU64, Ordering};
+#[cfg(target_has_atomic = "64")]
+use std::sync::atomic::AtomicU64;
+use std::sync::atomic::Ordering;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
--- a/bb8/src/internals.rs
+++ b/bb8/src/internals.rs
@@ -1,5 +1,3 @@
-#[cfg(not(target_has_atomic = "64"))]
-use portable_atomic::AtomicU64;
 use std::cmp::min;
 use std::collections::VecDeque;
 #[cfg(target_has_atomic = "64")]
@@ -8,6 +6,8 @@ use std::sync::atomic::Ordering;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
+#[cfg(not(target_has_atomic = "64"))]
+use portable_atomic::AtomicU64;
 use tokio::sync::Notify;
 
 use crate::api::{Builder, ManageConnection, QueueStrategy, State, Statistics};
--- a/bb8/Cargo.toml
+++ b/bb8/Cargo.toml
@@ -14,6 +14,9 @@ futures-util = { version = "0.3.2", default-features = false, features = ["alloc
 parking_lot = { version = "0.12", optional = true }
 tokio = { version = "1.0", features = ["rt", "sync", "time"] }
 
+[target.'cfg(not(target_has_atomic = "u64"))'.dependencies]
+portable-atomic = "1"
+
 [dev-dependencies]
 tokio = { version = "1.0", features = ["macros"] }
 


djc:
Why do you need support for 0.8.x?

flaviojs:
As stated in the issue, I am trying to compile sccache.
According to `cargo tree`, sccache depends on opendal, which depends on bb8.

OpenDAL has this in their Cargo.toml:
```toml
bb8 = { version = "0.8", optional = true }
```
Due to semver rules, the most recent v0.8.x version compatible with all dependencies will be pulled in.
Starting with v0.8.4 I am unable to compile bb8 due to missing AtomicU64, so I need a new v0.8.x version with this fix.

djc:
Maybe get OpenDAL to update to bb8 0.9 instead? As a single volunteer maintainer I don't have much bandwidth for maintaining older versions.

flaviojs:
According to https://crates.io/crates/bb8/reverse_dependencies a lot more crates would need updating...

This can be though of as a bug introduced in 0.8.4 that affects any system that does not have AtomicU64.
Ideally what needs to happen is:
 * merge this PR, bump master version to 0.9.1, and publish
 * create branch 0.8.x in the v0.8.6 tag. apply this PR to it, bump 0.8.x version to 0.8.7, publish and maybe yank 0.8.4 to 0.8.6 too(?)

I'm willing to do the work, but what I can do depends on the level of permissions I have.
Currently the most I can do is add a version bump 0.9.1 to this PR, and make a PR against a future 0.8.x branch with this fix and a version bump to 0.8.7.
If you are willing to trust a total stranger and give me whatever permissions are needed to do the rest, then I will do the rest. Note that I've never published a crate before so am not sure what is needed.

djc:
I'm definitely not giving a complete stranger a bunch of permissions. If this (arguably a pretty niche issue) is so important to you, let's discuss funding?

bb8/src/internals.rs
```diff
@@ -1,6 +1,10 @@
+#[cfg(not(target_has_atomic = "64"))]
+use portable_atomic::AtomicU64;
```
djc, line 2: Nit: this import should go in the next block, with external crate imports.

flaviojs: done

flaviojs:
Dude, I'm just a random programmer that found a bug, reported it to the appropriate place and provided a fix for it with the suggested dependency. Donating my free time **IS** the kind of "funding" a person like me provides.

If the funding you require is money then please wait for the unlikely case of someone whose job depends on this crate and who's company policy doesn't involve forking dependencies and fixing things internally... :disappointed:

djc:
> Dude, I'm just a random programmer that found a bug, reported it to the appropriate place and provided a fix for it with the suggested dependency. Donating my free time **IS** the kind of "funding" a person like me provides.
> 
> If the funding you require is money then please wait for the unlikely case of someone whose job depends on this crate and who's company policy doesn't involve forking dependencies and fixing things internally... 😞

I appreciate that it sucks to be in your position. However, as someone who provides hundreds of hours per month of free labor to a zillion companies, doing backports (for this kind of niche architecture) is where I draw the line.

(And after I pointed you in the right direction and reviewed your changes, I don't think I deserve being "Dude"d at.)

flaviojs:
I'm glad it got merged. It seems like 0.8.x won't be getting the fix which sucks a lot.
I'll inform opendal of the problem in this crate after you make a 0.9.x release with this fix.

Dude... you interpreted "dude" as something negative?
To avoid misunderstandings: dude is informal, I speak informally and directly pretty much all the time and I don't know of any other appropriate word for what I want to transmit from my native language.
Even if historically the work was bad, so far I have never actually seen a case where it was used in a negative way.

PR #242, opened 2025-02-17
redis: upgrade to redis 0.29

djc:


djc:
Published bb8-redis v0.21.0 at registry `crates-io`.

PR #244, opened 2025-02-27
add json feature

tuna2134:
Hello,

I am not sure if #183 is working, so I sent this Pull Requests. If you don't want it, you can close it.


djc:
I'd prefer to do handle all redis features at once (similar to #183) instead of doing one at a time. Are you up for that?

tuna2134:
Okay, I'll do it

PR #245, opened 2025-03-06
Replace mention of bb8-diesel with diesel_async in README.md

zekesonxx:
bb8-diesel (https://github.com/overdrivenpotato/bb8-diesel) hasn't been maintained in 3 years and doesn't support modern diesel without going to a random fork. diesel_async is well maintained, has been for several years, and supports bb8 as a backend.


djc:
Thanks!

PR #256, opened 2025-06-16
Redis 0.24

djc:


jplatte:
This PR title (and associated changelog entry) is pretty misleading. Sounds like this is the redis crate v0.24.

djc:
> This PR title (and associated changelog entry) is pretty misleading. Sounds like this is the redis crate v0.24.

Fixed.

PR #257, opened 2025-07-31
Bump Rust edition, version, and a few dependencies

maxnachlinger:


djc:
What is your motivation for this?

